PR #38, opened 2020-10-28
Improve and harden router deletion on agent

sebageek:
This is not yet a fix for the agent unscheduling problem, but this is at least half the way there.

### DIFF
--- a/asr1k_neutron_l3/common/cache_utils.py
+++ b/asr1k_neutron_l3/common/cache_utils.py
@@ -1,0 +1,58 @@
+# Copyright 2020 SAP SE
+#
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from neutron.common import cache_utils as neutron_cache_utils
+from oslo_log import log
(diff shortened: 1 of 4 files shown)


asr1k_neutron_l3/common/cache_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    key = _gen_cache_key("deleted-router", host, router_id)
+    cache.set(key, data)
+
+
+def get_deleted_router(host, router_id):
+    cache = get_cache()
+    if not cache:
+        return
+
+    key = _gen_cache_key("deleted-router", host, router_id)
```
joker-at-work: Would probably make sense to move this prefix into a variable as it's used in two places.

asr1k_neutron_l3/plugins/l3/rpc/rpc_api.py
```diff
@@ -55,11 +56,12 @@ def update_router_status(self, context, **kwargs):
...
-    def get_deleted_routers(self, context, **kwargs):
-        host = kwargs.get('host')
-        router_ids = kwargs.get('router_ids')
-
-        return self.l3plugin.get_sync_data(context, router_ids=router_ids, active=None, host=host)
+    def get_deleted_router(self, context, host, router_id):
+        router = cache_utils.get_deleted_router(host, router_id)
+        if not router:
+            LOG.debug("No entry for router %s on host %s in deleted router cache", router_id, host)
+            return None
```
joker-at-work: This `None` would not be necessary.

joker-at-work (commented):
In general, do we expect any problems with serializing the data of the router into the cache? Are there any objects, that aren't used too often, that might fail the serialization?

sebageek:
From what I've seen the output of `get_sync_data()` only contains dicts, lists, strings and numbers, so serializing should not be a problem. I didn't follow all code paths though. On the other hand this datastructure at least gets serialized for transport through the rpc framework from neutron to the asr1k agent.

joker-at-work (approved):
If it's going over the wire, then neutron probably removed all the problematic objects or has compatibility shims for serialization. :+1:

PR #52, opened 2021-06-07
Disable NTP servers on customer reachable interfaces

swagner-de:
Customers should not be able to consume NTP from this router. Hence, BD interfaces get the corresponding config option to disable the NTP server.


sebageek (approved):
Please also cherry-pick this to `stable/ussuri-m3`.

PR #58, opened 2021-07-08
Make sure eventlet is patched on ncc connect

sebageek:
With Neutron Ussuri we sometimes get a RuntimeError from eventlet that
suggests a double read on a socket is happening by different threads.
We want to make sure that when we use ncclient that eventlet is patched.
Therefore as a possible mitigation call again eventlet.monkey_patch(),
just if the asr1k driver was not used by first importing the
asr1k_l3_agent or asr1k_ml2_agent.


velp (approved):
👍  we must not forget to roll back these changes

sebageek:
monkey_patch() can be called multiple times to the patch shouldn't change anything in case it's already patched.

PR #59, opened 2021-07-09
Use patched ncclient for compatibility with eventlet

notandy:
the +sapcc version is build from https://github.com/sapcc/ncclient/tree/v0.6.12+sapcc and
is available via the inhouse pypi repository.


notandy:
I will enforce the version via upper-constraints, so pip will pick up the version it finds (+sapcc inside our CI), this obseletes this PR

PR #61, opened 2021-07-15
Minor brush-up for documentation

BarbarossaTM:
Fixed some typos while reading the docs and unified wording/spelling


BarbarossaTM:
Just learned, that `stable/ussuri-m3` would be better suited, so reopening the PR against that branch.

PR #63, opened 2021-07-21
Remove greenlet from asr1kl3 agent, use python threads instead

notandy:
This patch removes greenlet from asr1kl3 agent by stripping eventlet
(and thus monkey_patching), and replacing following (eventlet) based modules
with python-thread based:

neutron.Manager -> embedded into main class L3ASRAgentWithStateReport
oslo-service.looping_call -> futurist.periodics
oslo-service.service -> python Thread with graceful_exit signal handler
grenlet.greenpool -> futurist.ThreadPoolExecutor
neutron.rpc -> oslo_messaging low level api (to switch it to threading)

Why?
----
Removing eventlet and replacing it with python.threading has the huge benefit that
external libraries as ncclient doesn't need to be patched or prepared for eventlet.
For a single-process and single worker-queue based agent python threading performance
should be more than sufficient.

Fixes
-----
ncclient (upstream version, no more double socket exceptions)
graceful shutdown (all threads are gracefully stopped)

Reports
-------
Also, there is now a regular thread report that can be manually triggered via
`kill -SIGUSR2 <pid>`:

```
Currently running 8 periodic workers:
+----------------------------------------------------------------------------------------------------+--------+-------------+-----------+------+----------+-----------+----------------+-----------------+-------------------------+
|                                                Name                                                | Active | Periodicity |  Runs in  | Runs | Failures | Successes | Stop Requested | Average elapsed | Average elapsed waiting |
+----------------------------------------------------------------------------------------------------+--------+-------------+-----------+------+----------+-----------+----------------+-----------------+-------------------------+
|     asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent.periodic_requeue_routers_task     | False  |      60     |  30.6859s |  1   |    0     |     1     |     False      |     0.0002s     |         0.0007s         |
| asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent.periodic_refresh_address_scope_config | False  |      5      |  0.6895s  |  5   |    0     |     5     |     False      |     0.0621s     |         0.0002s         |
|        asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent.periodic_worker_report         | False  |     600     | 570.6783s |  0   |    0     |     0     |     False      |        .        |            .            |
|     asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgentWithStateReport._report_state      | False  |     30.0    |  0.6798s  |  0   |    0     |     0     |     False      |        .        |            .            |
|      asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent._periodic_sync_routers_task      | False  |     120     |  91.3323s |  0   |    0     |     0     |     False      |        .        |            .            |
|        asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent._periodic_scavenge_task        | False  |     120     |  91.3325s |  0   |    0     |     0     |     False      |        .        |            .            |
|         asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent._check_devices_alive          | False  |     60.0    |  31.3325s |  0   |    0     |     0     |     False      |        .        |            .            |
|             asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgent._clean_device             | False  |     600     | 571.7885s |  0   |    0     |     0     |     False      |        .        |            .            |
+----------------------------------------------------------------------------------------------------+--------+-------------+-----------+------+----------+-----------+----------------+-----------------+-------------------------+
Currently running 11 threads: ['MainThread', 'pydevd.Writer', 'pydevd.Reader', 'pydevd.CommandThread', 'pydevd.CheckAliveThread', 'Thread-7', 'ASR1k Worker Thread', 'Thread-8', 'Rabbit-heartbeat', 'Thread-10', 'Thread-11']
```

### DIFF
--- a/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
+++ b/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
@@ -268,7 +268,7 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
             try:
                 backdoor_socket_path = CONF.backdoor_socket.format(pid=os.getpid())
             except (KeyError, IndexError, ValueError) as e:
-                backdoor_socket_path = conf.backdoor_socket
+                backdoor_socket_path = CONF.backdoor_socket
                 LOG.warning("Could not apply format string to manhole "
                             "backdoor socket path ({}) - continuing with "
                             "unformatted path"
@@ -280,6 +280,7 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
 
         self.context = n_context.get_admin_context_without_session()
         self.plugin_rpc = L3PluginApi(topics.L3PLUGIN, host)
+        self.l3_agent_rpc = None
         self.fullsync = cfg.CONF.asr1k_l3.sync_active
         self.pause_process = False
         self.sync_routers_chunk_size = cfg.CONF.asr1k_l3.sync_chunk_size
@@ -287,9 +288,8 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
 
         self.asr1k_pair = asr1k_pair.ASR1KPair()
 
-        self._router_loop = None
-        self._stop = False
-        self.connection = None
+        self._router_loop_thread = None
+        self._router_loop_stop = False
         self._queue = asr1k_queue.RouterProcessingQueue()
         self._requeue = {}
         self._last_full_sync = timeutils.now()
@@ -358,20 +358,20 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
                 LOG.info("Orphan clean is active, starting cleaning loop")
                 self.worker.add(self._clean_device)
 
-            self._router_loop = threading.Thread(target=self._process_routers_loop, daemon=True)
-            self._router_loop.start()
+            self._router_loop_thread = threading.Thread(target=self._process_routers_loop, daemon=True)
+            self._router_loop_thread.start()
 
             LOG.info("L3 agent started")
 
     def exit_gracefully(self, *args):
-        self._stop = True
+        self._router_loop_stop = True
         self.worker.stop()
-        if self._router_loop:
-            self._router_loop.join()
-        if self.connection:
-            self.connection.stop()
+        if self._router_loop_thread:
+            self._router_loop_thread.join()
+        if self.l3_agent_rpc:
+            self.l3_agent_rpc.stop()
             # proccess all remaining messages
-            self.connection.wait()
+            self.l3_agent_rpc.wait()
 
     @periodics.periodic(CONF.asr1k.clean_orphan_interval)
     def _clean_device(self):
@@ -381,11 +381,11 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
         self.topic = topics.L3_AGENT
         self.endpoints = [self]
         target = oslo_messaging.Target(topic=self.topic, server=CONF.host)
-        self.connection = oslo_messaging.get_rpc_server(
+        self.l3_agent_rpc = oslo_messaging.get_rpc_server(
             n_rpc.TRANSPORT, target, self.endpoints,
             'threading', n_rpc.RequestContextSerializer(),
             access_policy=oslo_messaging.DefaultRPCAccessPolicy)
-        self.connection.start()
+        self.l3_agent_rpc.start()
 
     def trigger_sync(self, signum, frame):
         LOG.info("Setup full sync based on external signal")
@@ -734,7 +734,7 @@ class L3ASRAgent(operations.OperationsMixin, DeviceCleanerMixin):
                         "its now {}".format(poolsize))
 
         with ThreadPoolExecutor(max_workers=poolsize) as executer:
-            while not self._stop:
+            while not self._router_loop_stop:
                 # the executer worker queue is infinite and doesn't block
                 if self._queue.get_size():
                     executer.submit(self._process_router_update)
--- a/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
+++ b/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
@@ -15,11 +15,13 @@
 #    under the License.
 
 import datetime
+import os
 import threading
 import signal
 import sys
 import time
 
+import manhole
 from futurist import periodics, ThreadPoolExecutor
 from neutron.agent.common import resource_processing_queue as queue
 from neutron.agent.l3 import agent as l3_agent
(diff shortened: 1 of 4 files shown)


joker-at-work:
Will we install gdb with python extension by default, so we can get some kind of backdoor or is there a console-thread started, too?

notandy:
> Will we install gdb with python extension by default, so we can get some kind of backdoor or is there a console-thread started, too?

Good question, I don't think python provides something similar by default, maybe there is a similar library out there? I will look into installing gdb-python via loci pipeline, which also make sense for all non-eventlet openstack projects.

notandy:
https://pypi.org/project/manhole/ looks promising

notandy:
Just evaluated backdoor in qa-de-2:

```
> hammer k backdoor neutron-asr1k-agent-01-5b68cb88fc-qsm65
Select a container:

  ▶ neutron-asr1k
    neutron-asr1k-ml2
/var/lib/neutron/eventlet_backdoor-1.socket
Calling  /usr/bin/socat -d -d TCP-LISTEN:33442,bind=localhost unix-connect:/var/lib/neutron/eventlet_backdoor-1.socket
Executing rlwrap --histsize 2147483647 socat stdio TCP:localhost:54985

######### ProcessID=1, ThreadID=139641907050240 #########
... (manhole prints a stacktrace of all threads on startup)


Python 3.6.9 (default, Jan 26 2021, 15:33:00)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(ManholeConsole)
>>>
```

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -301,6 +287,9 @@ def __init__(self, host, conf=None):
 
         self.asr1k_pair = asr1k_pair.ASR1KPair()
 
+        self._router_loop = None
+        self._stop = False
+        self.connection = None
```
velp: I would prefer to use a more clear naming here like `self.rpc_conn` or `self.rpc_connection` because it was not really clear what kind of connection it is.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -430,10 +426,12 @@ def router_added_to_agent(self, context, payload):
         LOG.debug('Got router added to agent :%r', payload)
         self.routers_updated(context, payload)
 
-    def _check_devices_alive(self, context):
-        device_info = self.plugin_rpc.get_device_info(context)
+    @periodics.periodic(CONF.asr1k_l3.sync_interval / 2)
```
velp, line 429: Should we have configurable option for `CONF.asr1k_l3.sync_interval / 2` or it's already tested (and the best) value?

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -724,9 +733,14 @@ def _process_routers_loop(self):
...
-        pool = eventlet.GreenPool(size=poolsize)
-        while True:
-            pool.spawn_n(self._process_router_update)
+        with ThreadPoolExecutor(max_workers=poolsize) as executer:
+            while not self._stop:
+                # the executer worker queue is infinite and doesn't block
+                if self._queue.get_size():
+                    executer.submit(self._process_router_update)
+                else:
+                    time.sleep(5)
```
velp, line 742: `5` looks like some magic value. Should we add some comments here?

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -724,9 +733,14 @@ def _process_routers_loop(self):
             LOG.warning("The processing thread pool size has been reduced to match 'yang_connection_pool_size' "
                         "its now {}".format(poolsize))
 
-        pool = eventlet.GreenPool(size=poolsize)
-        while True:
-            pool.spawn_n(self._process_router_update)
+        with ThreadPoolExecutor(max_workers=poolsize) as executer:
```
velp, line 736: Creating a thread pool every N minutes is much more resource-intensive than creating green threads. Perhaps we should think about creating a thread pool and reusing it.

velp (commented):
It looks good generally, but there are a few notes on the readability of the code.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
notandy: good point, i will also harmonize the use of _ or without.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
notandy: This is a 1:1 adaption of the FixedIntervallLoopingCall: https://github.com/sapcc/asr1k-neutron-l3/pull/63/files/68f196a7594f61601095a9d85a2d95209802437e#diff-348264b14216d69b76bb1dfff7cb6bce5fc70667db9bb0e4b2126d77d77fe901L363

I didn't wanted to change any behaviour of the agent, this patch doesn't change anything apart removing eventlet.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
notandy: Yes, it's very magic. I already commented above that the worker queue is infinite, so we just wait a bit before checking for new queue items. The values is a quite arbitrary small value. Any suggestion?

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
notandy: Where do you see that a thread pool is created every N minutes, and where do you take these N minutes?

This threadpool is created exactly once. And this doesn't even mean that the threads of the pool are created - they are spun up as demand rises (and jobs submitted to the executer).

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
velp: ah, ok

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
velp: I did not understand the logic below. Question resolved.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
velp: Let's leave this for now. I don't think that we will need to control this period. in the near future.

velp (dismissed):
LGTM

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -338,63 +327,70 @@ def __init__(self, host, conf=None):
         #             continue
         #     break
         self.monitor = self._initialize_monitor()
+        self.worker = periodics.PeriodicWorker([
```
joker-at-work, line 330: Why don't you use [`PeriodicWorker.create()`](https://github.com/openstack/futurist/blob/d17f58d1bf26fde903578260c87da44c191258ae/futurist/periodics.py#L413) here, but instantiate the worker directly?

Wouldn't it make sense to create a worker with the `ThreadPoolExecutor` instead of the default `SynchronousExecutor`? As I understand it, the previous periodic-tasks would be able to run in parallel so this would better mimic previous behavior.

joker-at-work: Looking further into it, `neutron.service.Service` starts 2 loopingcalls in `start()`: one for reporting state and one for running the periodic tasks synchronous. This makes sure that we always report state, even if some periodic tasks take some time. Therefore, the default `SynchronousExecutor` might be fine here, but we cannot have the `self._report_state()` handled by `self.worker`, too.

PR #67, opened 2021-10-18
AZ aware router scheduling

sebageek:
When a router is scheduled with an AZ hint it will only be scheduled
onto an asr1k agent with that given AZ, otherwise it will be scheduled
onto an agent without an AZ set. None and 'nova' (the default AZ) are
both regarded as "no AZ set". Only a single AZ hint is allowed per
router.

A router now also checks if a network has an AZ set on router attach. If
the network has a hint but the router does not then association is
denied.

Agents now properly advertise their given AZ. The ML2 agent does not use
this information at the moment, as for ml2/portbinding AZ restrictions
will be enforced by a driver above the asr1k driver.

WIP, as I still need to do a proper codereview and also the flags to
declaw this are missing (i.e. we have the code ready, but not the
hardware and suddenly everything with AZ hint starts failing to get
deployed).

### DIFF
--- a/asr1k_neutron_l3/common/asr1k_constants.py
+++ b/asr1k_neutron_l3/common/asr1k_constants.py
@@ -32,3 +32,5 @@ ROUTER_STATE_ERROR = 'ERROR'
 
 SNAT_MODE_POOL = 'pool'
 SNAT_MODE_INTERFACE = 'interface'
+
+NO_AZ_LIST = (None, 'nova')
--- a/asr1k_neutron_l3/common/config.py
+++ b/asr1k_neutron_l3/common/config.py
@@ -44,6 +44,13 @@ ASR1K_OPTS = [
     cfg.BoolOpt('trace_yang_call_failures', default=False,
                 help=_("Log all failed YANG xml calls, including how long they took")),
 
+    cfg.BoolOpt('ignore_invalid_az_hint_for_router', default=False,
+                help="Router AZ hints will not be validated. This means that an AZ hint is accepted even if no asr1k "
+                     "agent is present with this AZ. This will also disable proper scheduling. ALL routers will "
+                     "be scheduled to any router of the driver's liking if no candidate with a matching AZ hint "
+                     "is present."),
+    cfg.BoolOpt('ignore_router_network_az_hint_mismatch', default=False,
+                help="Do not abort operation if router and network AZ hint do not match."),
 ]
 
 ASR1K_L3_OPTS = [
@@ -95,13 +102,6 @@ AGENT_STATE_OPTS = [
                 help=_('Log agent heartbeats')),
 ]
 
-AVAILABILITY_ZONE_OPTS = [
-    # The default AZ name "nova" is selected to match the default
-    # AZ name in Nova and Cinder.
-    cfg.StrOpt('availability_zone', max_length=255, default='nova',
-               help=_("Availability zone of this node")),
-]
-
 
 def create_device_pair_dictionary():
     device_dict = {}
@@ -139,19 +139,20 @@ def create_address_scope_dict():
     return address_scope_dict
 
 
-def register_l3_opts():
+def register_common_opts():
     cfg.CONF.register_opts(ASR1K_OPTS, "asr1k")
+    cfg.CONF.register_opts(AGENT_STATE_OPTS, 'AGENT')
+    common.register_availability_zone_opts_helper(cfg.CONF)
+
+
+def register_l3_opts():
     cfg.CONF.register_opts(ASR1K_L3_OPTS, "asr1k_l3")
     cfg.CONF.register_opts(ASR1K_L2_OPTS, "asr1k_l2")
-    cfg.CONF.register_opts(AGENT_STATE_OPTS, 'AGENT')
-    cfg.CONF.register_opts(AVAILABILITY_ZONE_OPTS, 'AGENT')
     common.register_interface_opts()
     common.register_interface_driver_opts_helper(cfg.CONF)
 
 
 def register_l2_opts():
-    cfg.CONF.register_opts(AGENT_STATE_OPTS, 'AGENT')
-    cfg.CONF.register_opts(ASR1K_OPTS, "asr1k")
     cfg.CONF.register_opts(ASR1K_L2_OPTS, "asr1k_l2")
     cfg.CONF.register_opts(service.RPC_EXTRA_OPTS)
 
--- a/asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
+++ b/asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
@@ -16,16 +16,24 @@
 
 
 import six
-from oslo_log import helpers as log_helpers
-from oslo_log import log as logging
 
-from asr1k_neutron_l3.common import asr1k_constants as constants
 from neutron.db import agentschedulers_db
+from neutron.db import agents_db as n_agents_db
+from neutron.db.availability_zone import router as router_az_db
 from neutron.db import l3_agentschedulers_db
 from neutron.db.models import agent as agent_model
 from neutron.extensions import l3agentscheduler
 from neutron.extensions import router_availability_zone as router_az
+from neutron_lib.db import api as db_api
 from neutron_lib.exceptions import agent as n_agent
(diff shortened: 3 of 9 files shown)


asr1k_neutron_l3/common/asr1k_constants.py
```diff
@@ -32,3 +32,5 @@
 
 SNAT_MODE_POOL = 'pool'
 SNAT_MODE_INTERFACE = 'interface'
+
+NO_AZ_LIST = (None, 'nova')
```
swagner-de, line 36: Wouldn't this make more sense make this one a `set()`

asr1k_neutron_l3/common/config.py
```diff
@@ -139,20 +139,23 @@ def create_address_scope_dict():
...
     common.register_interface_driver_opts_helper(cfg.CONF)
+    register_common_agent_opts()
 
 
 def register_l2_opts():
-    cfg.CONF.register_opts(AGENT_STATE_OPTS, 'AGENT')
-    cfg.CONF.register_opts(ASR1K_OPTS, "asr1k")
     cfg.CONF.register_opts(ASR1K_L2_OPTS, "asr1k_l2")
     cfg.CONF.register_opts(service.RPC_EXTRA_OPTS)
+    register_common_agent_opts()
```
swagner-de: Wouldn't it be cleaner to call `register_common_agent_opts` on ml2 agent startup as well as on l3 agent start instead of hijacking the the respective agent specific methods?

asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
```diff
@@ -116,8 +125,33 @@ def get_l3_agents(self, context, active=None, filters=None):
...
+    @db_api.retry_if_session_inactive()
+    def validate_availability_zones(self, context, resource_type,
+                                    availability_zones):
+        """Verify that the availability zones exist."""
+        if not availability_zones:
+            return
+        if len(availability_zones) > 1:
+            raise asr1k_exc.OnlyOneAZHintAllowed()
+        if resource_type == 'network':
+            agent_type = nl_const.AGENT_TYPE_DHCP
```
swagner-de: Is this code necessary here? We wouldn't do anything with networks?

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -43,12 +45,27 @@ def _get_candidates(self, plugin, context, sync_router):
                            'agent_id': current_l3_agents[0]['id']})
                 return []
 
-            candidates = plugin.get_l3_agents(context, active=True)
+            orig_candidates = plugin.get_l3_agents(context, active=True)
+
+            # router creation with az hint: only schedule on agent with appropriate AZ
+            # router creation without az hint: only schedule on agent with no AZ
+            az_hints = orig_az_hints = self._get_az_hints(sync_router)
```
swagner-de, line 52: TIL!

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -43,12 +45,27 @@ def _get_candidates(self, plugin, context, sync_router):
...
+            if not az_hints or az_hints[0] in asr1k_const.NO_AZ_LIST:
+                az_hints = asr1k_const.NO_AZ_LIST
+            candidates = [c for c in orig_candidates if c.availability_zone in az_hints]
+
+            if not candidates and orig_az_hints and cfg.CONF.asr1k.ignore_invalid_az_hint_for_router:
```
swagner-de, line 57: `_get_az_hints` returns either the az or the default config option. So bool(`orig_az_hints`) will always be true, right?

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -43,12 +45,27 @@ def _get_candidates(self, plugin, context, sync_router):
...
+                LOG.warning("No candidate found for router %s with az hint %s, reverting to original candidate list %s",
+                            sync_router['id'], az_hints, orig_candidates)
+                candidates = orig_candidates
+
             if not candidates:
-                LOG.warning('No active L3 agents')
+                LOG.warning('No active L3 agents found for router %s (az hints were %s)',
+                            sync_router['id'], orig_az_hints)
```
swagner-de, line 64: I assume the router precommit hook executed before scheduling? Otherwise we silently omit scheduling which I don't think is necessarily what we want.

asr1k_neutron_l3/common/asr1k_constants.py
sebageek: Set or tuple... I won't be modifying it and it's a two element thing, so I don't think this makes much of a difference speedwise, except for that the tuple would be immutable (which I like). Do you think a set makes more sense?

asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
sebageek: This is the method I'm overriding, verbatim copied and modified so it does the same thing it did before, so I didn't think much about if the code is needed, I wanted to stay close to the original. Original can be found here:

https://github.com/sapcc/neutron/blob/d6ba5ad3c418fa5ce99c7317966b66d0c525be4e/neutron/db/agents_db.py#L163-L179

Thinking about it we probably don't need to do any network validation, so I could remove those two lines. What do you think is better in this case?

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
sebageek: This is the implementation of `_get_az_hints()`: https://github.com/sapcc/neutron/blob/d6ba5ad3c418fa5ce99c7317966b66d0c525be4e/neutron/scheduler/l3_agent_scheduler.py#L368-L370

so it either returns the router's at hints or the default az, which in our infra is empty:
```
>>> from oslo_config import cfg
>>> cfg.CONF.default_availability_zones
[]
```
`bool([])` would then be False.

swagner-de: For some reason i thought that setting defaults to 'nova'.

asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
swagner-de: Yes. I noticed that and felt a bit grumpy we couldn't just call the parent method and check the exception. I'd keep it as small as possible but that's your choice.

asr1k_neutron_l3/common/asr1k_constants.py
swagner-de: Well we do not need the ordered property of a tuple and sets bring some nice set ops that I feel are handy often. I agree, we probably not utilize them in this case.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
sebageek: I think it does, I think I even checked this, but didn't do it conciously enough to be a 100% right now. I don't think we silently omit scheduling, the router will not be scheduled and I think it will be reported back to the user.

asr1k_neutron_l3/common/config.py
sebageek: Hmm I mean, `register_l3_opts()` and `register_l2_opts()` register everything you need, either for l3 and for l2... Not entirely sure if it makes sense to do an extra call to the common stuff.

But we could debate if this should really be named `register_common_agent_opts()`, as it also registers the `asr1k` opts, which we now also use in the server part.

asr1k_neutron_l3/plugins/l3/schedulers/asr1k_scheduler_db.py
sebageek: I'll remove the network part.

swagner-de (approved):

PR #68, opened 2021-10-25
Fix: Providing useful error messages in exceptions

rhalina:
Since Python 2.6 BaseException does not have an attribute `message` anymore. After the migration to Python 3.x the exceptions handled in `asr1koperations.py` didn't return a useful error message. This patch replaces `e.message` with `str(e)` to return the actual error.


sebageek (commented):
I like the explanation you put into the PR, could you please put that into your commit message as well?

Also, I think the same pattern exists in `asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py` in `_get_all_router_ids()`, could you fix it there as well as you're already at it?

sebageek (approved):

PR #69, opened 2021-10-28
Allow to overwrite default route via extra routes

swagner-de:
If a customer in the past configured a default route it was up to the
yang stack which route would be programmed, the openstack one pointing
to the real northbound gateway or the customers overwrite. This allows
explicitly overwritten default routes and removes the openstack default
route from the yang request.
Once a default route was overwritten, outbound connectivity for all
networks on that router will be broken, hence we set a tag indicating to
skip the monitoring. The namespace prober will be extended to make use
of this tag.

### DIFF
--- a/asr1k_neutron_l3/models/neutron/l3/router.py
+++ b/asr1k_neutron_l3/models/neutron/l3/router.py
@@ -150,16 +150,21 @@ class Router(Base):
 
     def _build_routes(self):
         routes = route.RouteCollection(self.router_id)
-        primary_route = self._primary_route()
-        if primary_route is not None and self._route_has_connected_interface(primary_route):
-            routes.append(primary_route)
 
+        # In case the customer sets a default route, we will restrain from programming the openstack primary route
+        primary_overridden = False
         for l3_route in self.router_info.get('routes', []):
             ip, netmask = utils.from_cidr(l3_route.get('destination'))
+            if netmask == '0.0.0.0':
+                primary_overridden = True
 
             r = route.Route(self.router_id, ip, netmask, l3_route.get('nexthop'))
             if self._route_has_connected_interface(r):
                 routes.append(r)
+
+        primary_route = self._primary_route()
+        if not primary_overridden and primary_route is not None and self._route_has_connected_interface(primary_route):
+            routes.append(primary_route)
 
         return routes
 
--- a/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
+++ b/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
@@ -19,6 +19,7 @@ import time
 from neutron.db import dns_db
 from neutron.db import extraroute_db
 from neutron.db import l3_gwmode_db as l3_db
+from neutron.extensions.tagging import TAG_PLUGIN_TYPE
 from neutron_lib.api.definitions import availability_zone as az_def
 from neutron_lib.callbacks import events
 from neutron_lib.callbacks import registry
@@ -354,10 +355,29 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
         asr1k_db.RouterAttsDb.ensure(context, result.get('id'))
         return result
 
+    def ensure_default_route_skip_monitoring(self, context, router_id, router):
+        tag_plugin = directory.get_plugin(TAG_PLUGIN_TYPE)
+        custom_default_route_tags = {constants.TAG_SKIP_MONITORING, constants.TAG_DEFAULT_ROUTE_OVERWRITE}
+
+        has_custom_default_route = any(r['destination'].endswith('/0') for r in router['routes'])
+        with context.session.begin(subtransactions=True):
+            router_tags = set(tag_plugin.get_tags(context, 'routers', router_id)['tags'])
+
+            has_default_route_tags = custom_default_route_tags.issubset(router_tags)
+
+            if has_default_route_tags != has_custom_default_route:
+                if has_default_route_tags:
+                    router_tags = router_tags - custom_default_route_tags
+                else:
+                    router_tags = router_tags.union(custom_default_route_tags)
+
+                tag_plugin.update_tags(context, 'routers', router_id, dict(tags=router_tags))
+
     @log_helpers.log_method_call
     def update_router(self, context, id, router):
         result = super(ASR1KPluginBase, self).update_router(context, id, router)
         asr1k_db.RouterAttsDb.ensure(context, result.get('id'))
+        self.ensure_default_route_skip_monitoring(context, id, result)
         return result
 
     @log_helpers.log_method_call
--- a/asr1k_neutron_l3/common/asr1k_constants.py
+++ b/asr1k_neutron_l3/common/asr1k_constants.py
@@ -34,3 +34,6 @@ SNAT_MODE_POOL = 'pool'
 SNAT_MODE_INTERFACE = 'interface'
 
 NO_AZ_LIST = (None, 'nova')
+
+TAG_DEFAULT_ROUTE_OVERWRITE = 'custom-default-route'
+TAG_SKIP_MONITORING = 'skip-monitoring'


asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -36,6 +38,9 @@
 
 LOG = log.getLogger(__name__)
 
+TAG_DEFAULT_ROUTE_OVERWRITE = 'custom-default-route'
+TAG_SKIP_MONITORING = 'skip-monitoring'
```
sebageek: Could be put into the `common/asr1k_constants.py`, maybe we want to use them somewhere else in some point in time?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -316,10 +321,29 @@ def create_router(self, context, router):
...
+        with context.session.begin(subtransactions=True):
+            router_tags = set(tag_plugin.get_tags(context, 'routers', router_id)['tags'])
+
+            has_default_route_tags = custom_default_route_tags.issubset(router_tags)
+
+            if has_default_route_tags != has_custom_default_route:
+                if has_default_route_tags:
+                    router_tags = router_tags - custom_default_route_tags
+                else:
+                    router_tags = router_tags.union(custom_default_route_tags)
```
sebageek, line 372: Okay, this took me a bit to understand. Essentialls this is:
```python
if has_default_route_tags and not has_custom_default_route:
    router_tags = router_tags - custom_default_route_tags
elif not has_default_route_tags and has_custom_default_route:
    router_tags = router_tags.union(custom_default_route_tags)
```
so with the tags and no default route, we remove both tags and with default route but no tags we add the tags, right?

Also, shorthand if you like
```python
router_tags |= custom_default_route_tags
```

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -316,10 +321,29 @@ def create_router(self, context, router):
         asr1k_db.RouterAttsDb.ensure(context, result.get('id'))
         return result
 
+    def ensure_default_route_skip_monitoring(self, context, router_id, router):
+        tag_plugin = directory.get_plugin(TAG_PLUGIN_TYPE)
+        custom_default_route_tags = {TAG_SKIP_MONITORING, TAG_DEFAULT_ROUTE_OVERWRITE}
+
+        has_custom_default_route = any(r['destination'] == '0.0.0.0/0' for r in router['routes'])
```
sebageek: I wondered if this might be different from the check we do above, as above we only check the netmask, here we check the subnet as well, but it's probably going to be hostbit-cleaned™ anyway, so we should be fine, right?

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -150,17 +150,22 @@ def get_internal_cidrs(self):
 
     def _build_routes(self):
         routes = route.RouteCollection(self.router_id)
-        primary_route = self._primary_route()
-        if primary_route is not None and self._route_has_connected_interface(primary_route):
-            routes.append(primary_route)
 
+        # In case the customer sets a default route, we will restrain from programming the openstack primary route
+        overwrite_primary = False
```
sebageek: I think I'd've named it `primary_overwritten`, as the customer decided to override the primary, not us. Also one could have [long discussions about if this should be overwrite or override](https://stackoverflow.com/questions/8651562/overwrite-or-override/8651622) (and I'd go with override), but it's nothing you need to change if you don't want to.

sebageek (commented):
Code looks good to me, some nitpick comments inside.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: Yes that makes sense.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: Not sure what you mean here. But `router_tags  |= custom_default_route_tags` is equal to `router_tags .update(custom_default_route_tags)` [(docs)](https://docs.python.org/3/library/stdtypes.html#set`). So that would only work for the second case. Was your intention to replace the whole block with one set operation?

I chose this way 2 level branch to disinguish clearly between _is there an asymmetry between routes and tags?_ and then _which way is the asymmetry?_

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: It typed as [AuthenticIPNetwork](https://github.com/sapcc/neutron/blob/stable/ussuri-m3/neutron/db/extraroute_db.py#L118) which in turn inherits from `netaddr.IPNetwork`. I would think it cleans the hostbit but it does not...
```
In [2]: from netaddr import IPNetwork

In [3]: str(IPNetwork('192.168.0.12/24'))
Out[3]: '192.168.0.12/24'
```
I will replace the check with `.endswith('/0')`

asr1k_neutron_l3/models/neutron/l3/router.py
swagner-de: Accepted.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Yes, `router_tags |= custom_default_route_tags` is equal to the statement in your code. I wanted to offer this as a shorthand, as I find it easier to parse in my head. `x |= y` tells me "add y to x" and the same would go for `router_tags -= custom_default_route_tags`.

Okay, I get why you chose to use the multilevel™ if-clause. For me the the variant of "if that is there and this is not do x; elif if this is there and that is not do y" is a bit more intuitive, but as both work, I'm fine with it.

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -150,17 +150,22 @@ def get_internal_cidrs(self):
...
+            if netmask == '0.0.0.0':
+                primary_overridden = True
 
             r = route.Route(self.router_id, ip, netmask, l3_route.get('nexthop'))
             if self._route_has_connected_interface(r):
                 routes.append(r)
 
+        primary_route = self._primary_route()
+        if not primary_overridden and primary_route is not None and self._route_has_connected_interface(primary_route):
+            routes.append(primary_route)
```
sebageek, line 167: I know this is a really unfun question, but... You're shifting the position of the `primary_route` in the list of routes for a VRF here. Semantically in the router config this should not make a difference, though when I look at our asr1k configs, the default route is always first in the list. Would this cause a reconfiguration of the list of routes in all VRFs that have more than a single default route?

Option 1: Live with that, it will happen once and probably not cause any problems
Option 2: Test what this actually does on device side, so we know what happens and deal with the result
Option 3: Use `routes.insert(0, primary_route)` to preserve the old order, whether we need it or not.

swagner-de: Let's try this out:
CLI config:
```
router#show run | in ip route vrf [redacted]
ip route vrf [redacted] 0.0.0.0 0.0.0.0 [redacted]
ip route vrf [redacted] 100.100.0.0 255.255.255.0 10.180.1.45
ip route vrf [redacted] 100.100.1.0 255.255.255.0 10.180.1.172
ip route vrf [redacted] 100.100.2.0 255.255.255.0 10.180.1.245
ip route vrf [redacted] 100.100.3.0 255.255.255.0 10.180.1.235
router#
```

YANG state of mind:
```
netconf-console --host router --get-config -N 'native=http://cisco.com/ns/yang/Cisco-IOS-XE-native' -x '/native/ip/route/vrf[name="[redacted]"]/ip-route-interface-forwarding-list'
<?xml version='1.0' encoding='UTF-8'?>
<data xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0">
  <native xmlns="http://cisco.com/ns/yang/Cisco-IOS-XE-native">
    <ip>
      <route>
        <vrf>
          <name>[redacted]</name>
          <ip-route-interface-forwarding-list>
            <prefix>0.0.0.0</prefix>
            <mask>0.0.0.0</mask>
            <fwd-list>
              <fwd>[redacted]</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.0.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.45</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.1.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.172</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.2.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.245</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.3.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.235</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
        </vrf>
      </route>
    </ip>
  </native>
</data>
```

Let's alter it:
```
~/devel ❯ cat edit-config.xml                                                           
  <native xmlns="http://cisco.com/ns/yang/Cisco-IOS-XE-native">
    <ip>
      <route>
        <vrf operation="merge">
          <name>[redacted]</name>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.0.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.45</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.1.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.172</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.2.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.245</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>100.100.3.0</prefix>
            <mask>255.255.255.0</mask>
            <fwd-list>
              <fwd>10.180.1.235</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
          <ip-route-interface-forwarding-list>
            <prefix>0.0.0.0</prefix>
            <mask>0.0.0.0</mask>
            <fwd-list>
              <fwd>[redacted]</fwd>
            </fwd-list>
          </ip-route-interface-forwarding-list>
        </vrf>
      </route>
    </ip>
  </native>%

~/devel ❯ netconf-console --host router  --edit-config edit-config.xml
<?xml version='1.0' encoding='UTF-8'?>
<ok xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0"/>
```

CLI again:
```
router#show run | in ip route vrf [redacted]
ip route vrf [redacted] 0.0.0.0 0.0.0.0 [redacted]
ip route vrf [redacted] 100.100.0.0 255.255.255.0 10.180.1.45
ip route vrf [redacted] 100.100.1.0 255.255.255.0 10.180.1.172
ip route vrf [redacted] 100.100.2.0 255.255.255.0 10.180.1.245
ip route vrf [redacted] 100.100.3.0 255.255.255.0 10.180.1.235
router#
```

-> Device does not care, nothing in the catchall log.

Now that I checked that I also think, the agent does notalter the sort-order either, cause we sort the routes by prefix when creating the YANG-dict-thing [here](https://github.com/sapcc/asr1k-neutron-l3/blob/stable/ussuri-m3/asr1k_neutron_l3/models/netconf_yang/route.py#L126).

sebageek: What a nice thing that we sort them, the commit message also explicitly mentions that :)

sebageek (approved):

PR #70, opened 2021-11-04
Apply different route-map on DAPnet extra routes

swagner-de:
Turns out _network devices_ that aggregate prefixes start to generate a
null0 static route as soon as any longer prefix than the aggregate is
present, even when a prefix with equal length as the aggregate is also
received via BGP the null0 route is still installed.

In Neutron, if a DAPnet enabled router chooses to add a static route
with a longer sized prefix, this prefix will lead to any BGP upstream
device installing the aggregates null0 route and the more specific
prefix for which a static route was just created. Due to the null0
route, connectivity will be lost to anything not contained in the static
route but contained in the larger null0 route.

To work around, we impose another route-map on static routes that pushes
a different community on the route. That community in turn we use again
to filter it out when announcing northbound to ACI.


sebageek (approved):
LGTM. I'll leave the grammar in the commit message and the extra whitespace in line 47 up to you.

PR #72, opened 2021-12-02
Throw API error if a networks segment exceeds the BD-VIF limit

swagner-de:
The ASR/Catalyst router platform has a limit that does not allow more
than 100 BD-VIFs per bridge-domain, or in our OpenStack setup more than
100 ports in an ASR bound segment. As we cannot work around that, we
will now raise an API exception and tell the user that we cannot assign
that network.

We also take care of this on scheduling in case some of the hardware
routers/agents are above limit while others are not.

### DIFF
--- a/asr1k_neutron_l3/common/config.py
+++ b/asr1k_neutron_l3/common/config.py
@@ -92,7 +92,8 @@ ASR1K_L2_OPTS = [
     cfg.IntOpt('sync_chunk_size', default=10, help=_("Number of ports to process in on poll")),
     cfg.StrOpt('external_interface', default=('1'), help=('')),
     cfg.StrOpt('loopback_external_interface', default=('2'), help=('')),
-    cfg.StrOpt('loopback_internal_interface', default=('3'), help=(''))
+    cfg.StrOpt('loopback_internal_interface', default=('3'), help=('')),
+    cfg.IntOpt('bdvif_bd_limit', default=100, help='Hardware limit of BD-VIF per Bridge Domain'),
 ]
 
 AGENT_STATE_OPTS = [
--- a/asr1k_neutron_l3/plugins/db/asr1k_db.py
+++ b/asr1k_neutron_l3/plugins/db/asr1k_db.py
@@ -15,6 +15,7 @@
 #    under the License.
 
 import random
+from typing import Dict
 
 from neutron.db import address_scope_db
 from neutron.db import db_base_plugin_v2
@@ -94,6 +95,15 @@ class DBPlugin(db_base_plugin_v2.NeutronDbPluginV2,
             if entry.advertise_extra_routes:
                 return True
         return False
+
+    def get_network_port_count_per_agent(self, context: n_context.Context, network_id: str) -> Dict[str, int]:
+        query = context.session.query(asr1k_models.ASR1KExtraAttsModel.agent_host,
+                                      func.count(asr1k_models.ASR1KExtraAttsModel.port_id).label('port_count')) \
+            .join(segment_models.NetworkSegment,
+                  segment_models.NetworkSegment.id == asr1k_models.ASR1KExtraAttsModel.segment_id) \
+            .filter(segment_models.NetworkSegment.network_id == network_id) \
+            .group_by(asr1k_models.ASR1KExtraAttsModel.agent_host)
+        return {x.agent_host: x.port_count for x in query.all()}
 
     def ensure_snat_mode(self, context, port_id, mode):
         if port_id is None:
--- a/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
+++ b/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
@@ -64,6 +64,15 @@ class SimpleASR1KScheduler(l3_agent_scheduler.AZLeastRoutersScheduler):
                             sync_router['id'], orig_az_hints)
                 return []
 
+            # Make sure the candidates do not reach the platforms hardware limit of BD-VIFs per BD
+            external_network_id = sync_router['external_gateway_info']['network_id']
+            agents_port_count = plugin.db.get_network_port_count_per_agent(context, external_network_id)
+            candidates = [c for c in candidates if agents_port_count.get(c.host, 0) < cfg.CONF.asr1k_l2.bdvif_bd_limit]
+            if not candidates:
+                LOG.warning(f'No L3 agents available that satisfy the BD-VIF '
+                            f'hardware limit for network {external_network_id}')
+                return []
+
             LOG.info("Found following candidates for router %s: %s",
                      sync_router['id'], ", ".join(c.host for c in candidates))
 
--- a/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
+++ b/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
@@ -470,6 +470,32 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
 
         return dict(result)
 
+    @registry.receives(resources.ROUTER_GATEWAY, [events.BEFORE_CREATE])
+    @log_helpers.log_method_call
+    def ensure_bdvif_limit(self, resource: str, event: str, trigger, payload: events.DBEventPayload):
+        """
+        There is a hardware limit on ASR1K/C8500 that allows us to not bind more than X BD-VIFs in a BD.
+        As configuration would just silently fail, we try to catch the situation on API level.
+        """
+        context = payload.context
+        network_id = payload.metadata['network_id']
+        router_id = payload.resource_id
+
+        host = self.get_l3_agents_hosting_routers(context, [router_id])
+        if len(host) == 0:
+            host = None
+        else:
+            host = host[0].host
+
+        port_count = self.db.get_network_port_count_per_agent(context, network_id)
+
+        # Either the router is scheduled and the agent has no room for BD-VIFs in that BD
+        # or it is not scheduled and all existing agents are above the limit
+        limit = cfg.CONF.asr1k_l2.bdvif_bd_limit
+        if (host and port_count[host] >= limit) \
+            or (not host and all(x >= limit for x in port_count.values())):
+            raise asr1k_exc.BdVifInBdExhausted(network_id=network_id, router_id=router_id)
+
     def ensure_config(self, context, id):
         asr1k_db.RouterAttsDb.ensure(context, id)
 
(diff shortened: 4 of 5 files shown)


asr1k_neutron_l3/common/config.py
```diff
@@ -92,7 +92,8 @@
     cfg.IntOpt('sync_chunk_size', default=10, help=_("Number of ports to process in on poll")),
     cfg.StrOpt('external_interface', default=('1'), help=('')),
     cfg.StrOpt('loopback_external_interface', default=('2'), help=('')),
-    cfg.StrOpt('loopback_internal_interface', default=('3'), help=(''))
+    cfg.StrOpt('loopback_internal_interface', default=('3'), help=('')),
+    cfg.IntOpt('bdvif_bd_limit', default=100, help='Hardware limit of BD-VIF per Bridge Domain')
```
sebageek: If you add a `,` here the next config option we add will not show up with the last config option in a diff ;)

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -95,6 +96,15 @@ def get_bgpvpn_advertise_extra_routes_by_router_id(self, context, router_id):
                 return True
         return False
 
+    def get_port_count_in_segment(self, context: n_context.Context, network_id: str) -> Dict[str, int]:
+        query = context.session.query(asr1k_models.ASR1KExtraAttsModel.agent_host,
+                                      func.count(asr1k_models.ASR1KExtraAttsModel.port_id).label('port_count')) \
+            .join(segment_models.NetworkSegment,
+                  segment_models.NetworkSegment.id == asr1k_models.ASR1KExtraAttsModel.segment_id) \
+            .filter(segment_models.NetworkSegment.network_id == network_id) \
+            .group_by(segment_models.NetworkSegment.id, asr1k_models.ASR1KExtraAttsModel.agent_host)
```
sebageek: I'm not exactly sure what difference it makes to group only by agent and to group by `networkSegment.id` as well. I haven't found a case where this makes a difference, but I'm not a 100% sure what I'm looking for. Do you have some insights on this? Else I'll continue trying to find this out.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -64,6 +64,18 @@ def _get_candidates(self, plugin, context, sync_router):
...
+            # Make sure the candidates do not reach the platforms hardware limit of BD-VIFs per BD
+            external_network_id = sync_router['external_gateway_info']['network_id']
+            agents_port_count = plugin.db.get_port_count_in_segment(context, external_network_id)
+            agent_hosts_under_limit = set([agent for agent, count in
+                                         agents_port_count.items()
+                                         if count < cfg.CONF.asr1k_l2.bdvif_bd_limit])
+            candidates = [c for c in candidates if c.host in agent_hosts_under_limit]
+            if not candidates:
+                LOG.warning(f'No L3 agents available that satisfy the BD-VIF '
+                            f'hardware limit for network {external_network_id}')
```
sebageek, line 73: Is this a problem we as operators would need to take care of immediately or will we have extra checks for this? If it's something we should notice, we could think of raising this to error level.

On a nitpicky sidenote, you could use `LOG` formatting, i.e. `LOG.warning("foo %s", external_network_id)`, which in theory is more performant, as the string formatting is only done when the respective loglevel is active.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -470,6 +470,32 @@ def get_config(self, context, id):
...
+            host = None
+        else:
+            host = host[0].host
+
+        port_count = self.db.get_port_count_in_segment(context, network_id)
+
+        # Either the router is scheduled and the agent has no room for BD-VIFs in that BD
+        # or it is not scheduled and all existing agents are above the limit
+        limit = cfg.CONF.asr1k_l2.bdvif_bd_limit
+        if (host and port_count[host] >= limit) \
```
sebageek, line 495: Was the limit of `100` still okay or was this already a problem? If it wasn't a problem, do you plan on keeping one as a spare?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -470,6 +470,32 @@ def get_config(self, context, id):
...
+            or (not host and all(x >= limit for x in port_count.values())):
```
sebageek, line 496: With line-length 120 this could be one line.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
sebageek: On a second review, it looks like this should in many cases already be caught by `ensure_bdvif_limit()` and an error should be shown to the user, right?

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -64,6 +64,18 @@ def _get_candidates(self, plugin, context, sync_router):
                             sync_router['id'], orig_az_hints)
                 return []
 
+            # Make sure the candidates do not reach the platforms hardware limit of BD-VIFs per BD
+            external_network_id = sync_router['external_gateway_info']['network_id']
+            agents_port_count = plugin.db.get_port_count_in_segment(context, external_network_id)
+            agent_hosts_under_limit = set([agent for agent, count in
+                                         agents_port_count.items()
+                                         if count < cfg.CONF.asr1k_l2.bdvif_bd_limit])
+            candidates = [c for c in candidates if c.host in agent_hosts_under_limit]
```
sebageek: If an external network is not already present on an asr1k agent (e.g. a new external network) then the agent will not appear as a key returned by `get_port_count_in_segment()` and thus never appear in `agent_hosts_under_limit`. How about this?
```python
candidates = [c for c in candidates if agents_port_count.get(c.host, 0) < cfg.CONF.asr1k_l2.bdvif_bd_limit]
```

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
swagner-de: Right. We have the hook that catches this on assignment of a gateway interface.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: I think 100 is stil ok. Why do you mean I keep 1 as a spare? It rejects the interface assignment if there are already 100 or more ports bound in that segment.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: I kept that in a separate line for readability.

asr1k_neutron_l3/plugins/db/asr1k_db.py
swagner-de: Back when I wrote that, my mind thought that in theory we could have multiple physnets to the same agent. Like one on Port-channel1 and another one on Port-channel2. In that case that grouping would make sense. Right?

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
swagner-de: That's shorter/better anyway. Thanks.

asr1k_neutron_l3/common/config.py
swagner-de: ok.

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -95,6 +96,15 @@ def get_bgpvpn_advertise_extra_routes_by_router_id(self, context, router_id):
                 return True
         return False
 
+    def get_port_count_in_segment(self, context: n_context.Context, network_id: str) -> Dict[str, int]:
```
sebageek: Maybe we should rename this method to `get_network_port_count_per_agent()` or something like that? I mean, we pass it a`network_id` and nothing is really "per segment".

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: physnets on port-channels don't make sense to me, I think these objects are in different domains.

In _theory_ each agent should be only mapped to one hostgroup in our ml2 driver, meaning each agent should have - with a consistent db - only one segment should exists per agent per network, everything else should be a misconfiguration on our part, I think.
If we have the case that this happens that we have two segments for the same network mapped to the same agent, then the query would return two lines for one agent, e.g. `(agent1, seg1, count1), (agent1, seg2, count2)`. In the dictionary, this would result in two cases for agent1 and one would overwrite the other like this:
```
>>> {x: n for n, x in enumerate((1,2,1,3,1))}
{1: 4, 2: 1, 3: 3}
```

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
sebageek: Okay. Do you want to keep the f-string in `LOG()`? If so, you can just mark this conversation as resolved.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: As you said, it rejects the assignment for 100 or more. If 100 is okay we could still allow 100, but the code is written so that 100 is not allowed. Just wanted to know if this is intentional or an off by one - or if I misunderstood something again.

sebageek: Aye, tnx fer explaining off-github. If we are at 100 we are already full, of course.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
swagner-de: Yes, I see the performance implication as neglectable.

asr1k_neutron_l3/plugins/db/asr1k_db.py
swagner-de: Okay, I see that possibility. The query is adjusted and does not rely on segments any more.

sebageek (approved):

PR #74, opened 2022-01-20
Support snat_enabled wit hl3-ext-gw-mode

sebageek:
By showing that our router plugin supports the l3-ext-gw-mode we can
allow the user to change snat_enabled attribute on a router. If we don't
advertise this as supported the user will just get an error:

Invalid input for external_gateway_info. Reason: Unexpected keys
supplied: enable_snat.


sebageek:
Not required in our cloud

PR #75, opened 2022-01-21
Fix real TypeError this time not KeyError

swagner-de:
This is a fix for a fix made in
433691a.
Instead fixing the TypeError, I really fixed a KeyError, even though the
code was trying to index into a NoneType.

### DIFF
--- a/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
+++ b/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
@@ -65,8 +65,9 @@ class SimpleASR1KScheduler(l3_agent_scheduler.AZLeastRoutersScheduler):
                 return []
 
             # Make sure the candidates do not reach the platforms hardware limit of BD-VIFs per BD
-            external_network_id = sync_router['external_gateway_info'].get('network_id')
-            if external_network_id:  # Internal interface is added before an external interface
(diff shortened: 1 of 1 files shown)


asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -65,8 +65,9 @@ def _get_candidates(self, plugin, context, sync_router):
                 return []
 
             # Make sure the candidates do not reach the platforms hardware limit of BD-VIFs per BD
-            external_network_id = sync_router['external_gateway_info'].get('network_id')
-            if external_network_id:  # Internal interface is added before an external interface
+            # Internal interface could be added before an external interface so check for presence
+            if sync_router['external_gateway_info'] and sync_router['external_gateway_info']['network_id']:
+                external_network_id = sync_router['external_gateway_info'].get('network_id')
```
sebageek: Either you are sure there will always be a `network_id` in there, then don't use `get()` here, or you are not sure if there'll be a `network_id` in there, then use `get()` in the if-clause.

swagner-de: I am sure, there will always be a network id when there is a network. I forgot to change that occurence.

sebageek (approved):

PR #78, opened 2022-04-13
Insert ownership data of l3 interfaces into on-device config

swagner-de:
This is done in order to pass data onwards to our snmp-exporters.
Interface description gets exposed as ifAlias in snmp, hence we are able
to pick up this string again and parse it in prometheus, generating some
labels to filter monitoring data on.

### DIFF
--- a/asr1k_neutron_l3/models/neutron/l3/interface.py
+++ b/asr1k_neutron_l3/models/neutron/l3/interface.py
@@ -125,8 +125,12 @@ class GatewayInterface(Interface):
         super(GatewayInterface, self).__init__(router_id, router_port, extra_atts)
 
         self.nat_address = self._nat_address()
+        
+        # annotate details about the router to the interface description so this can be picked up by SNMP
+        description = (f'type:gw;router:{self.router_id};network:{self.router_port["network_id"]};'
+                       f'subnet:{self._primary_subnet_id}')
 
-        self._rest_definition = VBInterface(name=self.bridge_domain, description=self.router_id,
+        self._rest_definition = VBInterface(name=self.bridge_domain, description=description,
                                             mac_address=self.mac_address, mtu=self.mtu, vrf=self.vrf,
                                             ip_address=self.ip_address,
                                             secondary_ip_addresses=self.secondary_ip_addresses, nat_outside=True,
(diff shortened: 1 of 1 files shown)


asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -125,8 +125,12 @@ def __init__(self, router_id, router_port, extra_atts):
         super(GatewayInterface, self).__init__(router_id, router_port, extra_atts)
 
         self.nat_address = self._nat_address()
+        
+        # annotate details about the router to the interface description so this can be picked up by SNMP
+        description = (f'type:gw;router:{self.router_id};network:{self.router_port["network_id"]};'
+                       f'subnet:{self._primary_subnet_id}')
```
sebageek, line 131: If you want to use fstrings would you be so kind and drop all python versions <3.6 from `setup.cfg` and add 3.6 - 3.10 in an extra commit?

sebageek, line 131: Have you looked into what the maximum string length for an interface description is or tested it?

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -146,7 +150,12 @@ def _nat_address(self):
 class InternalInterface(Interface):
     def __init__(self, router_id, router_port, extra_atts):
         super(InternalInterface, self).__init__(router_id, router_port, extra_atts)
-        self._rest_definition = VBInterface(name=self.bridge_domain, description=self.router_id,
+
+        # annotate details about the router to the interface description so this can be picked up by SNMP
+        description = (f'type:internal;project:{self.router_port["project_id"]};router:{self.router_id};'
+                      f'network:{self.router_port["network_id"]};subnet:{self._primary_subnet_id}')
```
sebageek, line 156: Why is the project only in this one, not in the upper one?

asr1k_neutron_l3/models/neutron/l3/interface.py
swagner-de: I did, it's 200 chars, which we are not hitting with these static length strings:
```
router(config-if)#description ?
  LINE  Up to 200 characters describing this interface
```

asr1k_neutron_l3/models/neutron/l3/interface.py
swagner-de: I was expecting this one. External interfaces and as such also the configuration object they are mapped to, do not belong to a project.
So, of course I could get that value piped through to the Interface object and put it in the description but 1) I need to touch more code and 2) it's not really correct.

asr1k_neutron_l3/models/neutron/l3/interface.py
swagner-de: I opened PR https://github.com/sapcc/asr1k-neutron-l3/pull/79 for it.

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: passt scho

sebageek (approved):

PR #79, opened 2022-04-13
Drop python <=3.5 compatibility and add >=3.6

swagner-de:


setup.cfg
```diff
@@ -15,8 +15,11 @@ classifier =
     Programming Language :: Python
     Programming Language :: Python :: 2
     Programming Language :: Python :: 2.7
```
sebageek: maybe drop python2 as well?

sebageek (approved):

PR #84, opened 2022-11-24
Stateless NAT mandatory in 17.6, introduce no-alias NAT

swagner-de:


asr1k_neutron_l3/models/netconf_yang/nat.py
```diff
@@ -681,7 +681,7 @@ def to_single_dict(self, context):
         entry[NATConstants.MATCH_IN_VRF] = ""
 
         if not self.from_device:
-            if self.stateless and context.has_stateless_nat:
+            if context.version_min_17_6 or (self.stateless and context.has_stateless_nat):
```
sebageek, line 686: We explicitly had this feature switch to enable/disable stateless nat if anything goes wrong with this feature. Are we sure we don't need that switch anymore with the new firmware version? If so we should remove this switch completely after we're done with the migration. OR we reintroduce it as a feature switch once we migrated.

asr1k_neutron_l3/models/netconf_yang/nat.py
```diff
@@ -599,6 +600,7 @@ def __parameters__(cls):
             {'key': 'mapping_id'},
             {'key': 'match_in_vrf', 'yang-key': 'match-in-vrf', 'default': False},
             {'key': 'stateless', 'yang-type': YANG_TYPE.EMPTY, 'default': False},
+            {'key': 'no_alias', 'yang-type': YANG_TYPE.EMPTY, 'default': False}
```
sebageek: Please add a `,` at the end of the line

asr1k_neutron_l3/tests/unit/models/netconf_yang/test_parsing.py
```diff
@@ -283,3 +284,45 @@ def test_bgp_parsing(self):
...
+        </native>
+    </data>
+</rpc-reply>
+"""
+
+        context = FakeASR1KContext()
+        snl = StaticNatList.from_xml(xml, context)
+        self.assertEqual(1, len(snl.static_nats))
+        nat = snl.static_nats[0]
+        self.assertEqual('10.20.30.40', nat.local_ip,)
```
sebageek: What's with the extra `,`?

asr1k_neutron_l3/tests/unit/models/netconf_yang/test_parsing.py
```diff
@@ -283,3 +284,45 @@ def test_bgp_parsing(self):
...
+        self.assertEqual('10.10.10.10', nat.global_ip)
+        self.assertEqual('route-it-like-they-do-on-discovery-channel', nat.vrf)
+        self.assertIsNone(nat.redundancy)
+        self.assertIsNone(nat.mapping_id)
+        self.assertEqual(True, nat.match_in_vrf)
+        self.assertEqual(True, nat.stateless)
+        self.assertEqual(True, nat.no_alias)
```
sebageek: use `assertTrue()` for these

asr1k_neutron_l3/tests/unit/models/netconf_yang/test_serialization.py
```diff
@@ -0,0 +1,62 @@
...
+                            <nat-static-transport-list-with-vrf>
+                                <local-ip>10.10.23.12</local-ip>
+                                <global-ip>192.168.23.12</global-ip>
+                                <vrf>you-and-me-baby-aint-nothin-but-mammals</vrf>
+                                <match-in-vrf></match-in-vrf>
+                            </nat-static-transport-list-with-vrf>""")[NATConstants.TRANSPORT_LIST]
+
+        for k in expected_17_3.keys():
+            if expected_17_3[k] is None:
+                expected_17_3[k] = ''
```
sebageek: Nothing you need to change, but just brainstoming:
* you could leave out the `.keys()` cause it's implicitly given
* you could use `items()` and have less to write
```python
for k, v in expected_17_3.items():
    if v is None:
        expected_17_3[k] = ''
```

asr1k_neutron_l3/models/netconf_yang/nat.py
swagner-de: I am ok with either one, but for the migration case Martin and I agreed that the driver shall turn on stateless NAT regardless of the setting once it sees a 17.6 device so the migration can be done without a redeploy.

sebageek: Makes sense. I'd just like it if we had some way to better document this / clean up after this. It just feels like something we might run into later if we don't take care of it after the migration. But yeah, works for me.

sebageek (approved):

PR #85, opened 2023-01-31
[yoga] neutron callback system passing dbevent payload instead of context, router_id, network_id

sven-rosenzweig:
For ressource router_interface and event before_create the parameter (context, router_id, network_id) passed to the callback system changed to dbevent payload. [(Notify call)](https://github.com/sapcc/neutron/blob/f9070846ab151daf428167d0e521e3ba482b3355/neutron/db/l3_db.py#L817-L822)

### DIFF
--- a/.github/workflows/run-tox.yml
+++ b/.github/workflows/run-tox.yml
@@ -2,7 +2,7 @@ name: run-tox
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


sven-rosenzweig:
Not sure if a pull request is needed in this case. Since this is not my code base I decided to open one.

PR #89, opened 2023-03-09
Use payloads / publish for router callbacks

sebageek:
Neutron dropped registry.notify() ans migrated to using payloads and publish(), so we're doing the same. This change is very much oriented on what can be found in neutron in neutron/agent/l3/agent.py.


joker-at-work (approved):
Looks like in Neutron to me

PR #91, opened 2023-04-20
Add option only to disable scheduling on an agent

swagner-de:
admin-shutting an agent leads to the agent being completely vacated. However we have cases in which we want an agent to stay online, but prevent any new workloads to be scheduled on it. This can be a capacity consideration or because it is meant to be decomissioned.

### DIFF
--- a/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
+++ b/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
@@ -825,9 +825,12 @@ class L3ASRAgentWithStateReport(L3ASRAgent):
             'availability_zone': self.conf.AGENT.availability_zone,
             'topic': topics.L3_AGENT,
             'configurations': {
-                'log_agent_heartbeats': self.conf.AGENT.log_agent_heartbeats},
+                'log_agent_heartbeats': self.conf.AGENT.log_agent_heartbeats,
+                'scheduling_disabled': self.conf.AGENT.scheduling_disabled,
+            },
             'start_flag': True,
-            'agent_type': constants.AGENT_TYPE_ASR1K_L3}
+            'agent_type': constants.AGENT_TYPE_ASR1K_L3,
+        }
         report_interval = self.conf.AGENT.report_interval
 
         if report_interval:
--- a/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
+++ b/asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
@@ -18,6 +18,8 @@ from neutron.scheduler import l3_agent_scheduler
 from oslo_config import cfg
 from oslo_log import helpers as log_helpers
 from oslo_log import log as logging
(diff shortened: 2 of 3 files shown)


asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -47,17 +49,29 @@ def _get_candidates(self, plugin, context, sync_router):
 
             orig_candidates = plugin.get_l3_agents(context, active=True)
 
+            enabled_candidates = []
+            disabled_hosts = []
+            for candidate in orig_candidates:
+                if not jsonutils.loads(candidate.configurations).get('scheduling_disabled', False):
+                    enabled_candidates.append(candidate)
+                else:
+                    disabled_hosts.append(candidate.host)
```
sebageek, line 58: Thinking out loud, as we don't use `orig_candidates` afterwards we could just filter the list. Would keep the diff smaller. On the other hand my short attempt of implementing this didn't really look nicer. Oh well.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
```diff
@@ -47,17 +49,29 @@ def _get_candidates(self, plugin, context, sync_router):
...
+
+            if disabled_hosts:
+                LOG.debug(f'Ignoring agent hosts {", ".join(disabled_hosts)} for scheduling of {sync_router["id"]}, '
+                          'scheduling disabled')
```
sebageek: I'd use format strings here. The formatting is only done when the log level is low enough for it to print. It's good practice also for `error`, as I think sentry can use this to better aggregate on it. I'm not a 100% sure on it, though.

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -825,9 +825,12 @@ def __init__(self, host, conf=None):
             'availability_zone': self.conf.AGENT.availability_zone,
             'topic': topics.L3_AGENT,
             'configurations': {
-                'log_agent_heartbeats': self.conf.AGENT.log_agent_heartbeats},
+                'log_agent_heartbeats': self.conf.AGENT.log_agent_heartbeats,
+                'scheduling_disabled': self.conf.AGENT.scheduling_disabled,
+            },
             'start_flag': True,
-            'agent_type': constants.AGENT_TYPE_ASR1K_L3}
+            'agent_type': constants.AGENT_TYPE_ASR1K_L3
```
sebageek: A `,` at the end of this line would be nice.

asr1k_neutron_l3/plugins/l3/schedulers/simple_asr1k_scheduler.py
swagner-de: At first hand I continued naming it `orig_candiates` but felt that the naming was kind of misleading and that `enabled_candidates` would indicate something was done to this list.

sebageek (approved):

PR #92, opened 2023-08-16
Delay route configuration after interface configuration

swagner-de:
When a router/VRF is part of a BGPVPN, the BGPVPN portion is configured
first. If a BGPVPN happens to have a peer on the local agent (hardware
router) this peer VRF's routes are leaked into the to be configured VRF.
Due to the order of operation, this leaking happens before any
VRF-attached l3 interfaces are configured.

Now say, a customer also configured an extra route. That extra route
also happens to be contained in a leaked route and that extra routes's
next hop also happens to be another BD-VIF in a BGPVPN-peered-VRF that
is local on that hardware router.

Even though the to be configured VRF has no l3 interface yet, it already
has a BGP RIB path pointing to that next hop address of the extra route.
It also believes it's a local interface, which it is, but it also is
not, because that BD-VIF lives in another VRF.

The hardware router not caring about such technicalities, will refuse to
configure the extra route with `%Invalid next hop address (it's this
router)` and all subsequent configuration fails.

The fact that the next hop lives on the same route can however be
obscured if the next hop is directly connected. It becomes directly
connected when the BD-VIFs are configure prior. So let's do that.

OpenStack also prevents you from creating extra routes where the next
hop is not local to the virtual router, so I think all cases should be
covered. Even on interface removal, OpenStack checks if a route still
utilizes that interface.

See https://github.com/sapcc/neutron/blob/5fb5653ffe700a8f9fb0949d16ecbf0978b61fb6/neutron/db/extraroute_db.py#L74-L86
https://github.com/sapcc/neutron/blob/5fb5653ffe700a8f9fb0949d16ecbf0978b61fb6/neutron/db/extraroute_db.py#L154-L164

### DIFF
--- a/asr1k_neutron_l3/models/neutron/l3/router.py
+++ b/asr1k_neutron_l3/models/neutron/l3/router.py
@@ -336,7 +336,6 @@ class Router(Base):
             results.append(self.pbr_acl.update())
         # Working assumption is that any NAT mode migration is completed
 
-        results.append(self.routes.update())
         results.append(self.floating_ips.update())
         results.append(self.arp_entries.update())
 
@@ -344,6 +343,8 @@ class Router(Base):
         for interface in self.interfaces.all_interfaces:
             if not isinstance(interface, l3_interface.OrphanedInterface):
                 results.append(interface.update())
+
+        results.append(self.routes.update())
 
         # We don't remove NAT statement or pool if enabling/disabling snat - instead update ACL
         if self.gateway_interface is not None:


sebageek (commented):
Only comments regarding the commit message:

1. `[...]not local the virtual[...]` is missing a "to" I think
2. [optional] Openstack --> OpenStack - Brand awareness™
3. The links towards code lines will soon be out of date (as soon as we do the next update of our branch where code touches the extraroute_db.py). Would it maybe make sense to link to the HEAD commit of our branch or maybe link to the openstack/neutron repo, as they don't do force pushes on updates?
4. From the commit message I am not 100% sure, so let me ask: Is there an additional API precommit hook we should introduce so the customer cannot build broken routers with bgpvpn or is everything already there with the static routes check?

swagner-de:
1. done
2. done
3. done
4. With the route programming delayed, we will still program such _less optimal_ configuration. I have also thought about a hook, yet the hook would require to gather all routers that take part in a BGPVPN, get their interal_interfaces + extra routes, check for overlaps and then deny that configuration. Never the less, my faith tells me that a customer should be free to break their router in whatever way they want.  We should not do the customer's sanity check.

sebageek:
Regarding 3. now it's linked to HEAD of the branch, which still can change the next time someone touches the file. What I was thinking about is that we could (should?) include a link that will still be valid once somebody pushes new changes to head, i.e. reference stuff by commit id.

sebageek (approved):

PR #93, opened 2023-08-22
Add floating ip ARP cleaning sync loop to agent

sebageek:
The agent now as an ARP cleaning syncloop. By default every two minutes (though it is configurable). The agent fetches all fip --> router mac associations. We expect each fip - if we see it - to be on the mac address of the gateway interface of our router. Then, for each device on the agent, we fetch the ARP cache and look for IPs that we know about and are on the wrong mac. Everything wrong will then be cleared.

Fetching all ARP entries from a device takes about 1.2-4.0s in my tests, depending on the ARP table size on the device. Selecting only a subset of fields makes the query go something like 20-40% faster.

---

Some things are still left open for discussion:
 * Our fips are not VRF or address scope aware, do we need to match fip address scope to the VRFs address scopes to check if we can clear it? FIPs might be unique, but needs checking (don't remember of the top of my mind).
 * Is fetching the ARP table possible everywhere where we need it (should run through the regions and do a short check)
 * Do we need to add an API point to run ARP cleaning via it? Probably not with 2 minutes, but with a longer interval it might be interesting.
 * Currently we do "whole region" arp cleaning, cause it is quite fast, but if it's not.... Would it make sense to bump the syncloop up to ~15 minutes and do a check where we see when a FIP was used the last time and specifically clear it everywhere?
     * fetching the ARP table could be filtered by IP (I think)
     * extra bookkeeping needed, as we would need to remember when a FIP was last used. Might be doable via standard attributes, might not be.

### DIFF
--- a/asr1k_neutron_l3/common/prometheus_monitor.py
+++ b/asr1k_neutron_l3/common/prometheus_monitor.py
@@ -39,6 +39,7 @@ DETAIL_LABELS = ['host', 'device', 'entity', 'action']
 BASIC_LABELS = ['host']
 STATS_LABELS = ['host', 'status']
 DEVICE_ENTITY_COUNT_LABELS = ['host', 'device', 'entity']
+FIP_ON_WRONG_MAC_COUNT_LABELS = ['host', 'device', 'vrf']
 
 L2 = "l2"
 L3 = "l3"
@@ -87,6 +88,9 @@ class PrometheusMonitor(object):
 
         self._device_entity_count = Gauge('device_entity_count', 'Number of instances of an entity present on device',
                                           DEVICE_ENTITY_COUNT_LABELS, namespace=self.namespace)
+        self._fip_on_wrong_mac_count = Counter('fip_on_wrong_mac_count',
+                                               'Number of times a FIP was found on the wrong mac',
+                                               FIP_ON_WRONG_MAC_COUNT_LABELS, namespace=self.namespace)
 
         if self.type == L3:
             self._router_create_duration = Histogram("router_create_duration", "Router create duration in seconds",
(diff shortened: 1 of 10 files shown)


sebageek:
We now reduce the arp timeout for the asr1ks to something like 30 minutes by default. The device load has been deemed to be acceptable, also for the large prod regions. Clearing only a subset still shows to be only something like 10-40% faster, leaning on the 10-20% part. Technically, cleaning things by address scope would be the correct approach, but cleaning it everywhere should not cause problems for us - if it does, we can add this feature later.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -227,6 +230,11 @@ def to_dict(self, context):
             else:
                 vbi[L3Constants.NTP][xml_utils.OPERATION] = NC_OPERATION.REMOVE
 
+        if self.arp_timeout and int(self.arp_timeout) > 0:
+            vbi[L3Constants.ARP] = {L3Constants.TIMEOUT: int(self.arp_timeout)}
+        else:
```
swagner-de, line 235: Wouldn't this lead to us sending a `delete` everytime we sync an interface, even if arp timeout is not in use and has not been configured either? Would that be an issue?

asr1k_neutron_l3/common/prometheus_monitor.py
```diff
@@ -39,6 +39,7 @@
 BASIC_LABELS = ['host']
 STATS_LABELS = ['host', 'status']
 DEVICE_ENTITY_COUNT_LABELS = ['host', 'device', 'entity']
+FIP_ON_WRONG_MAC_COUNT_LABELS = ['host', 'device']
```
swagner-de: How about we add the VRF to that label? We have the info and we could tell the customer right away that we have seen things on their VRF.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: Yes, we'd always send the delete requests (or better: state description) with the configuration, but as in most cases this is already the current state, confd on the asr1k shouldn't issue a cli command for it, i.e. this shouldn't be a costly operation.

swagner-de (approved):

PR #97, opened 2023-09-25
Don't get subnet CIDR if not present

sebageek:
When instantiating our Router when the router has no subnet (or we somehow didn't find one, the router was not deleted... I am not quite sure on the properties triggering this) this might result in an AttributeError, where in _build_pbr_acl() the
self.gateway_interface.primary_subnet is None and therefore has no member .get(). To fix this we simply check if the given subnet is None beforehand.

### DIFF
--- a/asr1k_neutron_l3/models/neutron/l3/interface.py
+++ b/asr1k_neutron_l3/models/neutron/l3/interface.py
@@ -90,6 +90,7 @@ class Interface(base.Base):
         for subnet in self.router_port.get('subnets', []):
             if subnet.get('id') == self._primary_subnet_id:
                 return subnet
+        return None
 
     @property
     def subnets(self):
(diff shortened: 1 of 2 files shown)


asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -90,6 +90,7 @@ def _primary_subnet(self):
         for subnet in self.router_port.get('subnets', []):
             if subnet.get('id') == self._primary_subnet_id:
                 return subnet
+        return None
```
swagner-de, line 93: Wouldn't this return `None` either way?

swagner-de (approved):

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: I once read in a style guide that its good style that if your method returns anything that it should have a return case for every case, also for the implicit `return None`. The idea behind this is to either have methods that return nothing (`None`) and then have methods that remind you that they return something in every case.

In this case it felt a bit like we ignored the default case, so I added an explicit return. Not important to keep it, just a thought :)

PR #100, opened 2023-11-14
Fallback to native/version for capability check

sebageek:
Newer versions of the Cisco router firmware do not advertise their Cisco-IOS_XE-native version anymore, so we have to fallback onto another way of determining the firmware version. Asking for native/version returns the current firmware version (17.3, 17.4, ...), which we can currently use for all present feature switches. If this works out properly we could also use it as the main way to determine the feature set of a device, even let our YANG Models match themselves on versions.


sebageek:
Closed in favor of #101

PR #101, opened 2023-11-15
Replace version check with firmware version

sebageek:
We used to check for special features and versions by looking at the revision date of the YANG models, specifically the Cisco-IOS-XE-native model. Newer versions (e.g. 17.3) apparently don't advertise the revision dates of these models without extra queries, so this won't work. As most feature switches we require can be set by looking at the major/minor version of the current firmware. Therefore we now query for each connection what the version is via "native/version". As we might require more fine grained feature switches in the future I'll leave in the machinery to do capability checks based on YANG model versions.

---

This is an alternative to #100, where we go "all in" with the new version check. Probably a good choice.

### DIFF
--- a/asr1k_neutron_l3/models/asr1k_pair.py
+++ b/asr1k_neutron_l3/models/asr1k_pair.py
@@ -19,6 +19,7 @@ from oslo_config import cfg
 
 from asr1k_neutron_l3.common import config as asr1k_config
 from asr1k_neutron_l3.common.asr1k_exceptions import VersionInfoNotAvailable
+from asr1k_neutron_l3.models.netconf_yang import xml_utils
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


asr1k_neutron_l3/models/asr1k_pair.py
```diff
@@ -82,14 +83,27 @@ def _collect_version_info(self):
...
-            self._has_stateless_nat = connection.check_capability(module="Cisco-IOS-XE-nat",
-                                                                  min_revision="2020-11-01")
+            # newer images don't advertise all their capabilities, so we need to check the version
+            ver_xml_data = connection.xpath_get("native/version")
+            ver_data = xml_utils.XMLUtils.to_raw_json(ver_xml_data.xml)
+            try:
+                ver = ver_data['rpc-reply']['data']['native']['version'].split(".")
+            except KeyError as e:
+                LOG.error("Tried to fetch version for host %s, but couldn't parse the response: %s", self.host, e)
+                raise CapabilityNotFoundException(host=self.host, entity="native/version")
```
swagner-de: Should this throw `VersionInfoNotAvailable`?

swagner-de (changes requested):

swagner-de (approved):

PR #107, opened 2024-02-23
Dynamic nat pool

sebageek:


sebageek:
Still some TODOs:
 * one ValueError needs to be converted to a proper exceptions
 * couple of FIXMEs
 * testing in qa landscape to verify device config viability
 * write db migration
 * think about moving to stestr

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -582,6 +582,28 @@ def get_floating_ips_with_router_macs(self, context, fips=None, router_id=None):
 
         return {e.floating_ip_address: e.mac_address for e in query}
 
+    def ensure_router_atts(self, context, router_id):
+        with db_api.CONTEXT_WRITER.using(context):
```
swagner-de, line 565: It could  happen that the `rds_used` is dirty by the time we allocate the RD.  I suppose packing all of it in a transaction as the previous code did,  still does not help, or will it? We cannot lock what's not in the table except we lock the whole table.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -342,20 +342,18 @@ def _get_extra_atts(self, context, router_ids, host=None):
 
     def _get_router_atts(self, context, router_ids):
         router_atts = self.db.get_router_atts_for_routers(context, router_ids)
+        return {ra.router_id: ra for ra in router_atts}
 
-        return_dict = {}
 
-        for router_att in router_atts:
-            if return_dict.get(router_att.get('router_id')) is None:
-                return_dict[router_att.get('router_id')] = {}
```
swagner-de, line 350: Love these lines.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -342,20 +342,18 @@ def _get_extra_atts(self, context, router_ids, host=None):
...
 
-            return_dict[router_att.get('router_id')] = router_att
-        return return_dict
+    @registry.receives(resources.ROUTER, [events.PRECOMMIT_CREATE])
+    def allocate_router_atts_on_create(self, resource, event, trigger, payload):
+        LOG.debug("Allocating extra atts for router %s", payload.resource_id)
+        self.db.ensure_router_atts(payload.context, payload.resource_id)
 
     @log_helpers.log_method_call
     def create_router(self, context, router):
```
swagner-de, line 356: You could remove the whole function now.

asr1k_neutron_l3/models/netconf_yang/nat.py
```diff
@@ -126,26 +129,29 @@ def __init__(self, **kwargs):
...
-        pool[NATConstants.ID] = self.id
-
-        pool[NATConstants.START_ADDRESS] = self.start_address
-        pool[NATConstants.END_ADDRESS] = self.end_address
-        pool[NATConstants.NETMASK] = self.netmask
-
-        result = OrderedDict()
-        result[NATConstants.POOL] = pool
+        result = {
+            NATConstants.POOL: {
```
swagner-de, line 133: Is it in idea  to put the `NATConstants.POOL` pool in the `_wrapper_preamble()` and the `remove_wrapper()`?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -356,6 +356,114 @@ def create_router(self, context, router):
...
+                               request_body, router=None):
+        ext_ips = info.get('external_fixed_ips', []) if info else []
+        orig_router_atts = self.db.get_router_att(context, router_id)
+        dynamic_nat_pool = None
+        if len(ext_ips) > 1:
+            dynamic_nat_pool = self._handle_dynamic_nat_pool_allocation(context, router_id, info, ext_ips)
+
+        # router = router or self._get_router(context, router_id)
+        result = super()._update_router_gw_info(context, router_id, info, request_body, router)
+        if dynamic_nat_pool is not None or orig_router_atts.dynamic_nat_pool:
```
swagner-de, line 369: I don't understand why we need `orig_router_atts` here. Couldn't we just omit this call when `dynamic_nat_pool is not None`? Cause that is the one we are persisting.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -356,6 +356,114 @@ def create_router(self, context, router):
...
+        The caller will have to write back the NAT pool as router_att themselves
+        """
+        # someone is trying to add a gw with a nat pool (multiple ips)
+        if len(ext_ips) == 2:
+            raise asr1k_exc.DynamicNatPoolNeedsToHaveAtLeastTwoIPs()
+
+        has_ip_specified = False
+        has_no_ip_specified = False
+        alloc_subnet_id = None
+        for ip_def in ext_ips[:-1]:
```
swagner-de, line 386: Why `[:-1]`?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -356,6 +356,114 @@ def create_router(self, context, router):
...
+            # check if we have mixed ip/no-ip requests
+            if 'ip_address' in ip_def:
+                has_ip_specified = True
+            else:
+                has_no_ip_specified = True
```
swagner-de, line 391:
```
has_ip_specified = any('ip_address' in x for x in ext_ips)
has_no_ip_specified = any('ip_address' not in x for x in ext_ips)
```
?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -16,10 +16,12 @@
 from collections import OrderedDict
 import time
 
+import netaddr
 from neutron.db import dns_db
 from neutron.db import extraroute_db
 from neutron.db import l3_gwmode_db as l3_db
 from neutron.extensions.tagging import TAG_PLUGIN_TYPE
+from neutron.ipam import driver
```
swagner-de: I feel like `driver` is a bit of a generic import, would you mind importing it as `ipam_driver`

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -356,6 +356,114 @@ def create_router(self, context, router):
...
+            av_iter = iter(av_set)
+            result = [next(av_iter)]
+            for ip in av_iter:
+                if int(ip) - 1 != int(result[-1]):
+                    # non-consecutive IP
+                    result = [ip]
+                    continue
+                result.append(ip)
+                if len(result) == ip_count:
+                    pool = f"{result[0]}-{result[-1]}/{subnet['cidr'].split('/')[1]}"
```
swagner-de: Maybe I am missing something but what about
```
av_ranges = sorted(av_set.iter_ipranges(), key=lambda x: x.size)
for av_range in av_ranges:
    if av_range.size >= ip_count:
         pool = str(av_range) + '/' + subnet['cidr'].split('/')[1]
```
I think this would also give us less fragmenting as we are using the smallest available gap in contrast to the first encountered gap.

asr1k_neutron_l3/plugins/db/models.py
```diff
@@ -44,6 +44,7 @@ class ASR1KRouterAttsModel(model_base.BASEV2):
     router_id = sa.Column(sa.String(length=36), sa.ForeignKey('routers.id', ondelete='CASCADE'), nullable=False,
                           primary_key=True)
     rd = sa.Column(sa.Integer(), nullable=False)
+    dynamic_nat_pool = sa.Column(sa.String(length=83), nullable=True)
```
swagner-de, line 48: Comment about the value length here?
I understand max IPv6 length is 39 chars, x 2 = 78 + 1 char for `-` + 4 chars for CIDR notation e.g. `/128`

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: Figured it out by myself. Last IP is the interface IP.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -356,6 +356,114 @@ def create_router(self, context, router):
...
+        else:
+            # all ips are specified, check that they are consecutive
+            ips = [ip_def['ip_address'] for ip_def in ext_ips[:-1]]
+            ip = netaddr.IPAddress(ips[0])
+
+            for next_ip in ips[1:]:
+                next_ip = netaddr.IPAddress(next_ip)
+                if int(next_ip) - int(ip) != 1:
+                    raise asr1k_exc.DynamicNatPoolSpecifiedIpsNotConsecutivelyAscending(ip=ip, next_ip=next_ip)
+                ip = next_ip
```
swagner-de: What about
```
In [17]: ips = [f'10.10.10.{i}' for i in range(10)]

In [18]: netaddr.IPSet(ips).iscontiguous()
Out[18]: True

In [19]: ips = [f'10.10.10.{i}' for i in range(10)] + ['10.10.10.20']

In [20]: netaddr.IPSet(ips).iscontiguous()
Out[20]: False

```

We could then also allow the user to provide them in whatever order (except the Interface IP), as long as they form a continuous space.

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: `db_api.CONTEXT_WRITER` should already open a transaction, just as we had it before, so the code is more or less equivalent. I thought about doing some extra work here and then decided to just go with the "as-is" state, as I only wanted to refactor this method away from that class we had before.

We could annotate the method with a db_retry decorator, but I originally found this to be out of scope.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: At the moment this still serves the purpose of logging each router creation call through the logging decorator that we have. So far I thought it'd still be useful to our debugging endeavors to have the log line there.

asr1k_neutron_l3/models/netconf_yang/nat.py
sebageek: We could do that but I think it would also change the yang options. To be honest, when I saw that we already had NATPooling enabled and that this was apparently even partly working, I just went with what was there. I removed some of the OrderedDicts, as I found them to be unnecessary and harder to read, but stuck with everything else.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: If the router used to have a dynamic nat pool but does not have one after the gateway update we need to persist that by updating the pool to `None`.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: I like it! Though we still need to iterate over all entries to do the `subnet_id` check, so doing this as two extra generator expressions would result in extra iterations that we'd do later on anyway in the check-`for`.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
swagner-de: That makes sense.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Good idea, will adapt it.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: So far I saw no reason to allow the user to give the IPs to us unsorted, I'd rather have them specify the pool in order, as I'd expect they do if they know what they're doing. Also, we would have to do something to detect addresses specified multiple times, as the set would deduplicate them. Or is that already done by neutron?

swagner-de: I feel like neutron does no such thing in the `_update_router_gw_info()` methods. But you could easily compare `IPSet.size == len(ext_ip) - 1` to achieve that.

swagner-de: Doesn't matter, it was just a suggestion to save a few lines.

swagner-de (dismissed):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek:
```
@@ -413,14 +416,11 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
             return dynamic_nat_pool
         else:
             # all ips are specified, check that they are consecutive
-            ips = [ip_def['ip_address'] for ip_def in ext_ips[:-1]]
-            ip = netaddr.IPAddress(ips[0])
-
-            for next_ip in ips[1:]:
-                next_ip = netaddr.IPAddress(next_ip)
-                if int(next_ip) - int(ip) != 1:
-                    raise asr1k_exc.DynamicNatPoolSpecifiedIpsNotConsecutivelyAscending(ip=ip, next_ip=next_ip)
-                ip = next_ip
+            ips = sorted([ip_def['ip_address'] for ip_def in ext_ips[:-1]])
+            ipset = netaddr.IPSet(ips)
+            if not ipset.iscontiguous():
+                a, b, *_ = list(ipset.iter_ipranges())
+                raise asr1k_exc.DynamicNatPoolSpecifiedIpsNotConsecutivelyAscending(ip=a[-1], next_ip=b[0])

```
That'd be my offer. With that we also allow the user to specify IPs in a random order, as long as the last IP is the interface ip. Also needs one extra test and one to be adjusted. Dupliate IPs are already found by the neutron API and rejected, so no extra code needed for that. Sounds good?

swagner-de (approved):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Implemented as discussed

PR #112, opened 2024-06-05
Basic ipv6 support

sebageek:
There is still some testing to be done. Also, the tests are not yet complete

### DIFF
--- a/asr1k_neutron_l3/common/utils.py
+++ b/asr1k_neutron_l3/common/utils.py
@@ -18,7 +18,7 @@ import re
 import socket
 import struct
 
-from netaddr import IPNetwork, IPAddress
+from netaddr import AddrFormatError, IPNetwork, IPAddress
 from oslo_log import log as logging
 
 
@@ -132,6 +132,14 @@ def to_wildcard_mask(prefix_len):
 
 def ip_in_network(ip, net):
     return IPAddress(ip) in IPNetwork(net)
+
+
+def get_ip_version(ip):
+    try:
+        ip = IPAddress(ip)
+        return ip.version
+    except AddrFormatError:
+        return None
 
 
 def network_in_network(net, parent_net):
--- a/asr1k_neutron_l3/common/utils.py
+++ b/asr1k_neutron_l3/common/utils.py
@@ -136,7 +136,7 @@ def ip_in_network(ip, net):
 
 def get_ip_version(ip):
     try:
-        ip = IPAddress(ip)
+        ip = IPNetwork(ip)
         return ip.version
     except AddrFormatError:
         return None
--- a/asr1k_neutron_l3/models/netconf_yang/bgp.py
+++ b/asr1k_neutron_l3/models/netconf_yang/bgp.py
@@ -28,8 +28,9 @@ class BGPConstants(object):
     ID = "id"
     ADDRESS_FAMILY = "address-family"
     IPV4 = "ipv4"
-    IPV6 = "ipv6"
     IPV4_UNICAST = "ipv4-unicast"
+    IPV6 = "ipv6"
+    IPV6_UNICAST = "ipv6-unicast"
     WITH_VRF = "with-vrf"
     AF_NAME = "af-name"
     NAME = "name"
@@ -46,7 +47,94 @@ class BGPConstants(object):
     ROUTE_MAP = "route-map"
 
 
-class AddressFamily(NyBase):
+class AddressFamilyBase(NyBase):
+    ITEM_KEY = BGPConstants.VRF
+
+    ID_FILTER = None
+    VRF_XPATH_FILTER = None
+    LIST_KEY = None
+
+    @classmethod
+    def get_for_vrf(cls, context, asn=None, vrf=None):
+        return cls._get_all(context=context, xpath_filter=cls.VRF_XPATH_FILTER.format(asn=asn, vrf=vrf))
+
+    @classmethod
+    def get_primary_filter(cls, **kwargs):
+        return cls.ID_FILTER.format(asn=kwargs.get('asn'), vrf=kwargs.get('vrf'))
+
+    @classmethod
+    @execute_on_pair(return_raw=True)
+    def get(cls, vrf, asn, context):
+        return cls._get(vrf=vrf, asn=asn, context=context)
+
+    @classmethod
+    @execute_on_pair(return_raw=True)
+    def exists(cls, vrf, asn, context):
+        return cls._exists(vrf=vrf, asn=asn, context=context)
+
+    @classmethod
+    def remove_wrapper(cls, dict, context):
+        dict = cls._remove_base_wrapper(dict, context)
+
+        if dict is None:
+            return None
+
+        dict = dict.get(BGPConstants.ROUTER, dict)
+        dict = dict.get(BGPConstants.BGP, dict)
+        asn = dict.get("id")
+        dict = dict.get(BGPConstants.ADDRESS_FAMILY, dict)
+        dict = dict.get(BGPConstants.WITH_VRF, dict)
+        dict = dict.get(cls.LIST_KEY, dict)
+
+        dict.setdefault(BGPConstants.VRF, {})["id"] = asn
+
+        return dict
+
+    def _wrapper_preamble(self, dict, context):
+        af = OrderedDict()
+        af[BGPConstants.AF_NAME] = BGPConstants.UNICAST
+        af.update(dict)
+        result = {}
+        result[self.LIST_KEY] = af
+        result = {BGPConstants.WITH_VRF: result}
+
+        bgp = OrderedDict()
+        bgp[BGPConstants.ID] = self.asn
+        bgp[BGPConstants.ADDRESS_FAMILY] = result
+        bgp[xml_utils.NS] = xml_utils.NS_CISCO_BGP
+        result = {BGPConstants.BGP: bgp}
+        result = {BGPConstants.ROUTER: result}
+        return result
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        self.enable_bgp = kwargs.get('enable_bgp', False)
+        if self.asn is None:
+            self.asn = kwargs.get("asn", None)
+
+    def to_dict(self, context):
+        raise NotImplementedError
+
+    def to_delete_dict(self, context):
+        if not self.vrf:
+            return {}
+
+        return {
+            BGPConstants.VRF: {
+                BGPConstants.NAME: self.vrf,
+            }
+        }
+
+    def _delete_no_retry(self, context, *args, **kwargs):
+        # only delete if present on device
+        device_af = self._internal_get(context=context)
+        if device_af and device_af.vrf:
+            return super()._delete_no_retry(context, *args, **kwargs)
+        return None
+
+
+class AddressFamilyV4(AddressFamilyBase):
     ID_FILTER = """
         <native xmlns="http://cisco.com/ns/yang/Cisco-IOS-XE-native"
                   xmlns:ios-bgp="http://cisco.com/ns/yang/Cisco-IOS-XE-bgp">
@@ -71,113 +159,94 @@ class AddressFamily(NyBase):
     VRF_XPATH_FILTER = ("/native/router/bgp[id='{asn}']/address-family/with-vrf/"
                         "ipv4[af-name='unicast']/vrf[name='{vrf}']")
     LIST_KEY = BGPConstants.IPV4
-    ITEM_KEY = BGPConstants.VRF
-
-    @classmethod
-    def get_for_vrf(cls, context, asn=None, vrf=None):
-        return cls._get_all(context=context, xpath_filter=cls.VRF_XPATH_FILTER.format(asn=asn, vrf=vrf))
 
     @classmethod
     def __parameters__(cls):
         return [
             {'key': 'asn', 'id': True, 'yang-key': 'id'},
             {'key': 'vrf', 'yang-key': 'name'},
-            {'key': 'networks_v4', 'yang-path': 'ipv4-unicast/network', 'yang-key': BGPConstants.WITH_MASK,
-             'type': [Network], 'default': []},
+            {'key': 'networks', 'yang-path': 'ipv4-unicast/network', 'yang-key': BGPConstants.WITH_MASK,
+             'type': [NetworkV4], 'default': []},
         ]
 
-    @classmethod
-    def get_primary_filter(cls, **kwargs):
-        return cls.ID_FILTER.format(asn=kwargs.get('asn'), vrf=kwargs.get('vrf'))
+    def to_dict(self, context):
+        if self.vrf is None:
+            return {}
 
-    @classmethod
-    @execute_on_pair(return_raw=True)
-    def get(cls, vrf, asn, context):
-        return super(AddressFamily, cls)._get(vrf=vrf, asn=asn, context=context)
+        vrf = {
+            BGPConstants.NAME: self.vrf,
+            BGPConstants.IPV4_UNICAST: {
+                xml_utils.OPERATION: NC_OPERATION.PUT,
+                BGPConstants.NETWORK: {
+                    BGPConstants.WITH_MASK: [
+                        net.to_dict(context) for net in sorted(self.networks, key=lambda x: (x.number, x.mask))
+                    ],
+                },
+            }
+        }
 
-    @classmethod
-    @execute_on_pair(return_raw=True)
-    def exists(cls, vrf, asn, context):
-        return super(AddressFamily, cls)._exists(vrf=vrf, asn=asn, context=context)
-
-    @classmethod
-    def remove_wrapper(cls, dict, context):
-        dict = super(AddressFamily, cls)._remove_base_wrapper(dict, context)
-
-        if dict is not None:
-            dict = dict.get(BGPConstants.ROUTER, dict)
-            dict = dict.get(BGPConstants.BGP, dict)
-            asn = dict.get("id", None)
-            dict = dict.get(BGPConstants.ADDRESS_FAMILY, dict)
-            dict = dict.get(BGPConstants.WITH_VRF, dict)
-            dict = dict.get(BGPConstants.IPV4, dict)
-
-            if dict.get(BGPConstants.VRF, None) is not None:
-                dict[BGPConstants.VRF]["id"] = asn
-            else:
-                dict[BGPConstants.VRF] = OrderedDict()
-                dict[BGPConstants.VRF]["id"] = asn
-
-        return dict
-
-    def _wrapper_preamble(self, dict, context):
-        af = OrderedDict()
-        af[BGPConstants.AF_NAME] = BGPConstants.UNICAST
-        af.update(dict)
-        result = {}
-        result[self.LIST_KEY] = af
-        result = {BGPConstants.WITH_VRF: result}
-
-        bgp = OrderedDict()
-        bgp[BGPConstants.ID] = self.asn
-        bgp[BGPConstants.ADDRESS_FAMILY] = result
-        bgp[xml_utils.NS] = xml_utils.NS_CISCO_BGP
-        result = {BGPConstants.BGP: bgp}
-        result = {BGPConstants.ROUTER: result}
+        result = {
+            BGPConstants.VRF: vrf,
+        }
         return result
 
-    def __init__(self, **kwargs):
-        super(AddressFamily, self).__init__(**kwargs)
 
-        self.enable_bgp = kwargs.get('enable_bgp', False)
-        if self.asn is None:
-            self.asn = kwargs.get("asn", None)
+class AddressFamilyV6(AddressFamilyBase):
+    ID_FILTER = """
+        <native xmlns="http://cisco.com/ns/yang/Cisco-IOS-XE-native"
+                  xmlns:ios-bgp="http://cisco.com/ns/yang/Cisco-IOS-XE-bgp">
+            <router>
+                <ios-bgp:bgp>
+                    <ios-bgp:id>{asn}</ios-bgp:id>
+                    <ios-bgp:address-family>
+                        <ios-bgp:with-vrf>
+                            <ios-bgp:ipv6>
+                                <ios-bgp:af-name>unicast</ios-bgp:af-name>
+                                <ios-bgp:vrf>
+                                    <ios-bgp:name>{vrf}</ios-bgp:name>
+                                </ios-bgp:vrf>
+                            </ios-bgp:ipv6>
+                        </ios-bgp:with-vrf>
+                    </ios-bgp:address-family>
+                </ios-bgp:bgp>
+            </router>
+        </native>
+    """
+
+    VRF_XPATH_FILTER = ("/native/router/bgp[id='{asn}']/address-family/with-vrf/"
+                        "ipv6[af-name='unicast']/vrf[name='{vrf}']")
+    LIST_KEY = BGPConstants.IPV6
+
+    @classmethod
+    def __parameters__(cls):
+        return [
+            {'key': 'asn', 'id': True, 'yang-key': 'id'},
+            {'key': 'vrf', 'yang-key': 'name'},
+            {'key': 'networks', 'yang-path': BGPConstants.IPV6_UNICAST, 'yang-key': BGPConstants.NETWORK,
+             'type': [NetworkV6], 'default': []},
+        ]
 
     def to_dict(self, context):
-        result = OrderedDict()
-        if self.vrf is not None:
-            vrf = OrderedDict()
-            vrf[BGPConstants.NAME] = self.vrf
-            vrf[BGPConstants.IPV4_UNICAST] = {
+        if self.vrf is None:
+            return {}
+
+        vrf = {
+            BGPConstants.NAME: self.vrf,
+            BGPConstants.IPV6_UNICAST: {
                 xml_utils.OPERATION: NC_OPERATION.PUT,
+                BGPConstants.NETWORK: [
+                    net.to_dict(context) for net in sorted(self.networks, key=lambda x: x.number)
+                ],
             }
+        }
 
-            if self.networks_v4:
-                vrf[BGPConstants.IPV4_UNICAST][BGPConstants.NETWORK] = {
-                    BGPConstants.WITH_MASK: [
-                        net.to_dict(context) for net in sorted(self.networks_v4, key=lambda x: (x.number, x.mask))
-                    ]
-                }
-            result[BGPConstants.VRF] = vrf
-        return dict(result)
-
-    def to_delete_dict(self, context):
-        result = OrderedDict()
-        if self.vrf is not None:
-            vrf = OrderedDict()
-            vrf[BGPConstants.NAME] = self.vrf
-            result[BGPConstants.VRF] = vrf
-
-        return dict(result)
-
-    def _delete_no_retry(self, context, *args, **kwargs):
-        # only delete if present on device
-        device_af = self._internal_get(context=context)
-        if device_af and device_af.vrf:
-            return super(AddressFamily, self)._delete_no_retry(context, *args, **kwargs)
+        result = {
+            BGPConstants.VRF: vrf,
+        }
+        return result
 
 
-class Network(NyBase):
+class NetworkV4(NyBase):
     @classmethod
     def __parameters__(cls):
         return [
@@ -196,9 +265,33 @@ class Network(NyBase):
         return to_cidr(self.number, str(self.mask))
 
     def to_dict(self, context):
-        net = OrderedDict()
+        net = {}
         net[BGPConstants.NUMBER] = self.number
         net[BGPConstants.MASK] = self.mask
+        if self.route_map:
+            net[BGPConstants.ROUTE_MAP] = self.route_map
+        return net
+
+
+class NetworkV6(NyBase):
+    @classmethod
+    def __parameters__(cls):
+        return [
+            {'key': 'number'},
+            {'key': 'route_map', 'yang-key': BGPConstants.ROUTE_MAP, 'default': None},
+        ]
+
+    @classmethod
+    def from_cidr(cls, cidr, route_map=None):
+        return cls(number=cidr, route_map=route_map)
+
+    @property
+    def cidr(self):
+        return self.number
+
+    def to_dict(self, context):
+        net = {}
+        net[BGPConstants.NUMBER] = self.number.lower()
         if self.route_map:
             net[BGPConstants.ROUTE_MAP] = self.route_map
         return net
--- a/asr1k_neutron_l3/models/netconf_yang/bgp.py
+++ b/asr1k_neutron_l3/models/netconf_yang/bgp.py
@@ -37,6 +37,7 @@ class BGPConstants(object):
     VRF = "vrf"
     REDISTRIBUTE = "redistribute"
     REDISTRIBUTE_VRF = "redistribute-vrf"
+    REDISTRIBUTE_V6 = "redistribute-v6"
     CONNECTED = "connected"
     STATIC = "static"
     UNICAST = "unicast"
@@ -45,6 +46,7 @@ class BGPConstants(object):
     NUMBER = "number"
     MASK = "mask"
     ROUTE_MAP = "route-map"
+    DEFAULT = "default"
 
 
 class AddressFamilyBase(NyBase):
@@ -104,6 +106,18 @@ class AddressFamilyBase(NyBase):
         bgp[xml_utils.NS] = xml_utils.NS_CISCO_BGP
         result = {BGPConstants.BGP: bgp}
         result = {BGPConstants.ROUTER: result}
+        return result
+
+    @classmethod
+    def _get(cls, **kwargs):
+        # make sure the result has a vrf associated
+        # due to the way we filter for our AF (with the ASN included) we always get a result, which
+        # prompts the original _get() method to always return a class, which results in problems when
+        # we want an empty result in case it's not there (should_be_none=True), therefore we filter out
+        # pseudoempty results here
+        result = super()._get(**kwargs)
+        if result and getattr(result, "vrf", None) is None:
+            return None
         return result
 
     def __init__(self, **kwargs):
@@ -167,26 +181,38 @@ class AddressFamilyV4(AddressFamilyBase):
             {'key': 'vrf', 'yang-key': 'name'},
             {'key': 'networks', 'yang-path': 'ipv4-unicast/network', 'yang-key': BGPConstants.WITH_MASK,
              'type': [NetworkV4], 'default': []},
+            {'key': 'redistribute_connected_with_rm', 'yang-key': 'route-map',
+             'yang-path': 'ipv4-unicast/redistribute-vrf/connected'},
+            {'key': 'redistribute_static_with_rm', 'yang-key': 'route-map',
+             'yang-path': 'ipv4-unicast/redistribute-vrf/static/default'},
         ]
 
     def to_dict(self, context):
         if self.vrf is None:
             return {}
 
-        vrf = {
-            BGPConstants.NAME: self.vrf,
-            BGPConstants.IPV4_UNICAST: {
-                xml_utils.OPERATION: NC_OPERATION.PUT,
-                BGPConstants.NETWORK: {
-                    BGPConstants.WITH_MASK: [
-                        net.to_dict(context) for net in sorted(self.networks, key=lambda x: (x.number, x.mask))
-                    ],
-                },
-            }
+        ipv4_unicast = {
+            xml_utils.OPERATION: NC_OPERATION.PUT,
+            BGPConstants.NETWORK: {
+                BGPConstants.WITH_MASK: [
+                    net.to_dict(context) for net in sorted(self.networks, key=lambda x: (x.number, x.mask))
+                ],
+            },
         }
+        if self.redistribute_connected_with_rm or self.redistribute_static_with_rm:
+            redist = {}
+            if self.redistribute_connected_with_rm:
+                redist[BGPConstants.CONNECTED] = {BGPConstants.ROUTE_MAP: self.redistribute_connected_with_rm}
+            if self.redistribute_static_with_rm:
+                redist[BGPConstants.STATIC] = {
+                    BGPConstants.DEFAULT: {BGPConstants.ROUTE_MAP: self.redistribute_static_with_rm}}
+            ipv4_unicast[BGPConstants.REDISTRIBUTE_VRF] = redist
 
         result = {
-            BGPConstants.VRF: vrf,
+            BGPConstants.VRF: {
+                BGPConstants.NAME: self.vrf,
+                BGPConstants.IPV4_UNICAST: ipv4_unicast,
+            }
         }
         return result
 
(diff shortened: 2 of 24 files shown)


asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -233,16 +242,18 @@ def postflight(self, context, method):
 
         routes = []
         try:
-            routes = VrfRoute.get_for_vrf(context=context, vrf=self.id)
-            if len(routes) == 0:
-                LOG.info("No routes to clean")
+            routes_v4 = VrfRouteV4.get_for_vrf(context=context, vrf=self.id)
+            routes_v6 = VrfRouteV6.get_for_vrf(context=context, vrf=self.id)
+            routes = routes_v4 + routes_v6
+            if routes:
```
mutax: `if not routes` or `if routes`?

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -263,17 +274,18 @@ def postflight(self, context, method):
...
                 LOG.info("No address fammilies to clean")
 
             for af in afs:
-                LOG.info("Deleting hanging address family in vrf {} postflight.".format(self.name))
-                LOG.info(af)
+                LOG.info("Deleting hanging address family %s in vrf %s postflight.", af, self.name)
                 result = af._delete(context=context)
                 LOG.debug(result)
-                LOG.info("Deleted hanging address family in vrf {} postflight.".format(self.name))
+                LOG.debug("Deleted hanging address family in vrf %s postflight.", self.name)
```
mutax: add `af` here as well?

asr1k_neutron_l3/common/utils.py
```diff
@@ -134,6 +134,14 @@ def ip_in_network(ip, net):
     return IPAddress(ip) in IPNetwork(net)
 
 
+def get_ip_version(ip):
```
mutax, line 137: would rename to `get_ip_addr_version` or `get_ip_net_version`, or something along those lines,  get_ip_version sounds somehow generic to me.  (nitpick)

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -52,59 +52,114 @@ def all_interfaces(self):
...
+                for subnet in self.gateway_interface.subnets
+                if subnet.get('cidr') and utils.get_ip_version(subnet['cidr']) == ip_version]
+
+    def get_external_cidrs_v4(self):
+        return self.get_external_cidrs(4)
+
+    def get_external_cidrs_v6(self):
+        return self.get_external_cidrs(6)
+
+    def get_routable_networks(self, ip_version):
```
sven-rosenzweig, line 84: `_get_routable_networks` to mark those method private?

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -52,59 +52,114 @@ def all_interfaces(self):
...
+        if not self.gateway_interface:
+            return []
+
+        # routable networks are networks for which the subnet's subnet pool address scope
+        # matches the address scope of the external gateway
+        scope_key = {4: "address_scope_v4", 6: "address_scope_v6"}[ip_version]
```
sven-rosenzweig, line 90: Do we need to catch an `KeyError` here?

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -52,59 +52,114 @@ def all_interfaces(self):
...
+        return self.get_internal_cidrs(4)
+
+    def get_internal_cidrs_v6(self):
+        return self.get_internal_cidrs(6)
+
+    def get_external_cidrs(self, ip_version):
+        if not self.gateway_interface:
+            return []
+
+        return [subnet['cidr']
```
sven-rosenzweig, line 74: From my point of view this is easier to understand: 
```
 subnets = []
        for interface in self.internal_interfaces:
            for subnet in interface.subnets:
                if subnet.get('cidr') and utils.get_ip_version(subnet['cidr']) == ip_version:
                    subnets.append(subnet['cidr'])
```

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -52,59 +52,114 @@ def all_interfaces(self):
 
         return result + self.internal_interfaces + self.orphaned_interfaces
 
+    def get_internal_cidrs(self, ip_version):
```
sven-rosenzweig: ` def _get_internal_cidrs(self, ip_version):` mark this as private?

asr1k_neutron_l3/models/netconf_yang/bgp.py
```diff
@@ -44,79 +46,49 @@ class BGPConstants(object):
...
-            dict = dict.get(BGPConstants.ADDRESS_FAMILY, dict)
-            dict = dict.get(BGPConstants.WITH_VRF, dict)
-            dict = dict.get(BGPConstants.IPV4, dict)
-
-            if dict.get(BGPConstants.VRF, None) is not None:
-                dict[BGPConstants.VRF]["id"] = asn
-            else:
-                dict[BGPConstants.VRF] = OrderedDict()
-                dict[BGPConstants.VRF]["id"] = asn
+        dict = cls._remove_base_wrapper(dict, context)
```
mutax, line 79: Would it make sense to use this opportunity to rename the parameter from `dict` to something that does not conflict with an internal datatype? Each time I see this code again after some time,  I stop and need some cycles to process what's going on there. Also my IDE complains :)

asr1k_neutron_l3/models/netconf_yang/bgp.py
```diff
@@ -136,48 +108,181 @@ def _wrapper_preamble(self, dict, context):
...
+                BGPConstants.NAME: self.vrf,
+            }
+        }
 
     def _delete_no_retry(self, context, *args, **kwargs):
         # only delete if present on device
         device_af = self._internal_get(context=context)
         if device_af and device_af.vrf:
-            return super(AddressFamily, self)._delete_no_retry(context, *args, **kwargs)
+            return super()._delete_no_retry(context, *args, **kwargs)
```
mutax, line 147: nitpick: what do we think regarding an explicit `return None` in such cases?

asr1k_neutron_l3/models/netconf_yang/bgp.py
```diff
@@ -196,9 +301,33 @@ def cidr(self):
...
+            {'key': 'number'},
+            {'key': 'route_map', 'yang-key': BGPConstants.ROUTE_MAP, 'default': None},
+        ]
+
+    @classmethod
+    def from_cidr(cls, cidr, route_map=None):
+        return cls(number=cidr, route_map=route_map)
+
+    @property
+    def cidr(self):
```
mutax, line 326: would it make sense to add `from_cidr()` and `cidr()` to `class NetworkV4` as well, so they provide the same functionality? Maybe adding a common parent class NetworkBase, like with some of the others?
(nitpick)

asr1k_neutron_l3/models/netconf_yang/route.py
```diff
@@ -87,23 +90,21 @@ def get_primary_filter(cls, **kwargs):
 
     @classmethod
     def remove_wrapper(cls, dict, context):
-        dict = super(VrfRoute, cls)._remove_base_wrapper(dict, context)
+        dict = cls._remove_base_wrapper(dict, context)
```
mutax, line 70: same question again - can we rename the variable from dict to something else?

asr1k_neutron_l3/models/netconf_yang/route.py
```diff
@@ -87,23 +90,21 @@ def get_primary_filter(cls, **kwargs):
...
-        result = {}
-        result[self.LIST_KEY] = dict
-        result = {RouteConstants.IP: result}
-        return result
-
-    def __init__(self, **kwargs):
-        super(VrfRoute, self).__init__(**kwargs)
+        return {
+            self.IP_KEY: {
+                self.LIST_KEY: dict,
```
mutax, line 82: dict variable shadowing builtin (added note here for completeness)

asr1k_neutron_l3/models/neutron/l3/prefix.py
```diff
@@ -13,70 +13,79 @@
...
+        for n, pfx in enumerate(prefixes, 1):
+            self._rest_definition.add_seq(
+                self._make_seq(n * 10, pfx)
+            )
 
-        self._rest_definition = prefix.Prefix(name="{}-{}".format(name_prefix, self.vrf))
+        if add_deny_if_empty and not prefixes:
+            self._rest_definition.add_seq(
+                # use 4242 as a likely not-used seq no
+                self._make_seq(4242, self.DEFAULT, action="deny")
```
mutax, line 37: make 4242 a centrally defined constant?

asr1k_neutron_l3/models/neutron/l3/route.py
```diff
@@ -19,41 +19,61 @@
 from asr1k_neutron_l3.models.neutron.l3 import base
 
 
-class RouteCollection(base.Base):
+class BaseRouteCollection(base.Base):
```
mutax: I think with most (all?) of the newly introduced Base-classes for v4/v6, "Base" was added at the end of the existing class name, so to follow that "naming scheme" this class would be called `RouterCollectionBase` .  That also fits nicely with `RouteCollectionV{4,6}`

asr1k_neutron_l3/models/neutron/l3/route.py
```diff
@@ -19,41 +19,61 @@
...
+
+
+class RouteCollectionV6(BaseRouteCollection):
+    @property
+    def _route_collection_model(self):
+        return l3_route.VrfRouteV6
+
+
+class RouteV6(base.Base):
+    def __init__(self, router_id, destination, nexthop):
```
mutax, line 73: Would it be possible (and make sense) to have RouteV4 and RouteV6 accept the same parameters in their constructor? Then they could also get a common parent class and avoid a bit of code duplication.

asr1k_neutron_l3/models/neutron/l3/route_map.py
```diff
@@ -31,28 +31,29 @@ def __init__(self, name, rt=None, routable_interface=False):
...
                                                        prefix_list='route-{}'.format(self.vrf),
-                                                       asn=[self.secondary_rt, 'additive'],
-                                                       enable_bgp=self.enable_bgp))
+                                                       asn=[self.secondary_rt, 'additive']))
                 seq += 10
 
-        sequences.append(route_map.MapSequence(seq_no=seq, operation='deny', prefix_list='ext-{}'.format(self.vrf)))
+        sequences.append(route_map.MapSequence(seq_no=seq, operation='deny', prefix_list=f'ext-{self.vrf}'))
+
+        # ipv6
```
mutax, line 54: to follow the scheme I would suggest adding a seq+=10 in line 53 and move the one from line 55 to after the append, but within the if branch. (possible nitpick)

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -222,40 +225,53 @@ def _build_nat_acl(self):
...
-            for subnet in int_port.get('subnets', []):
-                if subnet and subnet.get('cidr') and utils.ip_in_network(l3_route.nexthop, subnet['cidr']):
-                    return True
-
-        return False
+        all_cidrs = (
+            self.interfaces.get_external_cidrs_v4() +
+            self.interfaces.get_external_cidrs_v6() +
+            self.interfaces.get_internal_cidrs_v4() +
+            self.interfaces.get_internal_cidrs_v4()
```
mutax: copy paste error, the last one should be v6?

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -297,17 +313,77 @@ def _build_prefix_lists(self):
...
-        no_snat_interfaces = self.address_scope_matches()
+        return result
 
-        result.append(prefix.SnatPrefix(router_id=self.router_id, gateway_interface=self.gateway_interface,
-                                        internal_interfaces=no_snat_interfaces))
+    def _build_route_maps(self):
+        extraroutes_rt = None
+        if self.rt and ':' in self.rt:
+            # 65126:106 --> 65126:1106
+            asn, sf = self.rt.split(":", 2)
```
mutax: Should the 2 be a 1?

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -297,17 +313,77 @@ def _build_prefix_lists(self):
...
+                               routable_interface=self.routable_interface, enable_ipv6=self.enable_ipv6),
 
-        return result
+            # pbr route-map
+            route_map.PBRRouteMap(self.router_id,
+                                  has_gateway_interface=bool(self.gateway_interface)),
+
+            # redistribute routemaps
+            route_map.Redist4RouteMap(self.router_id, self.rt, extraroutes_rt,
+                                      enabled=(self.enable_ipv4 and
```
mutax: I have seen `self.enable_ipv4 and cfg.CONF.asr1k_l3.advertise_bgp_ipv4_routes_via_redistribute` several times now, might improve readability and reduce change for mistakes to add a class variable or property for that? (nitpick)

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -459,21 +527,24 @@ def _delete(self):
...
-        results.append(self.route_map.delete())
+        for rm in self.route_maps:
+            results.append(rm.delete())
         results.append(self.floating_ips.delete())
         results.append(self.arp_entries.delete())
-        results.append(self.routes.delete())
+        results.append(self.routes[4].delete())
+        results.append(self.routes[6].delete())
 
         for key in self.dynamic_nat.keys():
```
mutax, line 539: could also iterate over `values()` and drop the `get()`

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -161,8 +165,12 @@ def to_dict(self, context):
         # hopefully
         definition[VrfConstants.RD] = self.rd
 
-        if self.address_family_ipv4 is not None:
-            definition[VrfConstants.ADDRESS_FAMILY][VrfConstants.IPV4] = self.address_family_ipv4.to_dict(context)
+        if self.address_family_ipv4 or self.address_family_ipv6:
```
sven-rosenzweig, line 173:
```
 af = definition[VrfConstants.ADDRESS_FAMILY] = {}
 if self.address_family_ipv4 is not None:
                af[VrfConstants.IPV4] = self.address_family_ipv4.to_dict(context)
 if self.address_family_ipv6 is not None:
                af[VrfConstants.IPV6] = ""
```

Why do we need the outer if-clause?

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -65,6 +66,7 @@ def __init__(self, router_id, router_port, extra_atts):
         self._primary_subnet_id = None
         self.ip_address = self._ip_address()
```
sven-rosenzweig: For consistency reasons, does it make sense to rename this to  `self.ip4_addresses`?

asr1k_neutron_l3/models/netconf_yang/route.py
```diff
@@ -68,7 +68,10 @@ class VrfRoute(NyBase):
     VRF_XPATH_FILTER = "/native/ip/route/vrf[name='{vrf}']"
```
sven-rosenzweig: We could set this variable to ``None`` as this is overrriten in ```VrfRouteV6 or VrfRouteV4`

asr1k_neutron_l3/models/netconf_yang/route.py
```diff
@@ -113,37 +114,41 @@ def neutron_router_id(self):
...
-        vrf_route = OrderedDict()
-        vrf_route[RouteConstants.NAME] = self.name
-        vrf_route[RouteConstants.FOWARDING] = []
+        if not self.routes:
+            # no routes --> empty container
+            return {}
 
-        if isinstance(self.routes, list):
+        vrf_routes = []
+        if self.routes:
```
sven-rosenzweig: Do we really need to check twice as we have``if not self.routes:``?

asr1k_neutron_l3/models/neutron/l3/prefix.py
```diff
@@ -32,48 +34,72 @@ def __init__(self, name_prefix, router_id, gateway_interface, internal_interface
...
     def __init__(self, router_id=None, gateway_interface=None, internal_interfaces=None):
-        super(ExtPrefix, self).__init__(name_prefix='ext', router_id=router_id, gateway_interface=gateway_interface,
-                                        internal_interfaces=internal_interfaces)
+        super().__init__(name_prefix='ext', router_id=router_id, gateway_interface=gateway_interface,
+                         internal_interfaces=internal_interfaces)
 
         if self.gateway_interface is not None:
             i = 1
             for subnet in sorted(self.gateway_interface.subnets, key=itemgetter('id')):
+                if utils.get_ip_version(subnet.get('cidr')) != self.IP_VERSION:
```
sven-rosenzweig: Does it make sense to check whether `self.IP_VERSION `is not ``None``?

asr1k_neutron_l3/common/utils.py
sebageek: Currently this method accepts both an ip network or address and regardless of what it is it'll return the version. I think I'd keep it, as both proposed names would suggest it wouldn't take the other format.

asr1k_neutron_l3/models/netconf_yang/bgp.py
sebageek: This opportunity... probably not. But this is also something that has annoyed me for a long time. If we do this, I'd like to do it for all of the `remove_wrapper()` methods at the same time - just in case something is using that as a keyword argument somewhere.

asr1k_neutron_l3/models/netconf_yang/bgp.py
sebageek: "If we start returning something, we should return something in every path". As I only changed the super(), I didn't touch anything else here. I'll add one in this case :)

asr1k_neutron_l3/models/netconf_yang/bgp.py
sebageek: I see a `from_cidr()` and `cidr()` for in my `NetworkV4` and I think they're indeed used in a way that they provide the same interface to the NetworkV4/V6 class. Introducing a common baseclass could be done to have a defined interface, but the methods have subtle differences, so we couldn't share the code.

asr1k_neutron_l3/models/netconf_yang/route.py
sebageek: Let's do it in a followup

asr1k_neutron_l3/models/netconf_yang/route.py
sebageek: ....no, that doesn't make sense at all.

asr1k_neutron_l3/models/netconf_yang/vrf.py
sebageek: We need to make sure not to override the whole address family dict, but we also only want to set the ADDRESS_FAMILY key if we have an address family to set. So first, we check if we have any, then create the entry and then fill that up with whatever address families are available. We had cases where we just had the inner ifs and if both were true we'd override the first entry with the second entry. In this case, we'd have a `definitionVrfConstants.ADDRESS_FAMILY] = ...` in both cases.

asr1k_neutron_l3/models/netconf_yang/vrf.py
sebageek: The former.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: Yes, but I'd move this to a follow up / code cleanup run. In this PR I only tried to clean up what was hindering me in developing the IPv6 feature.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: 100%, we always want to specify the amount of splits we wanna do, not the number of results we want :)

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: It's used in several places in the models.neutron.l3.router.Router class, where we have the ip version in a variable anyways. Why should I make it private?

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: Did you mean to reference L56-59 with the double-for in `get_internal_cidrs()`? If so, I can also roll this out, if this is easier to understand. I find double-fors in a list-comprehension come quite easy, once you roll them out in your head, but I agree that they might not be the most common thing to see. I can roll it out in this case if it helps the understandability and we'll discuss it on a case-by-case basis.

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: Used externally by Router class

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: We hardcode the 4 and 6 everywhere (or loop over a hardcoded list) so I think this is fine. If it breaks, we need to fix it, so handling this would be "look into sentry, fix it in the code". I don't  think we need to handle this case here by catching that exception.

asr1k_neutron_l3/models/neutron/l3/prefix.py
sebageek: Hmm I don't see a benefit in it yet. We don't use it anywhere else. I think I'd keep it as-is for now.

asr1k_neutron_l3/models/neutron/l3/route.py
sebageek: Incorporated that, as discussed in Slack

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: Yes, most definitely! Good find!

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: This is the only place where this is used. The other two places have `ip_version` in a local variable and don't depend on `self.enable_ipv4` (in some cases the class itself has a `delete()` hook if the AF is disabled and therefore still needs to be created). So sadly I can't easily summarize them.

asr1k_neutron_l3/models/neutron/l3/interface.py
mutax: ..until IPv8 comes along 😸

asr1k_neutron_l3/models/neutron/l3/nat.py
```diff
@@ -53,23 +53,20 @@ def _rest_definition(self):
...
         self.mode = mode
 
         self.id = utils.vrf_to_access_list_id(self.router_id)
-        self.bridge_domain = bridge_domain
+        self.bridge_domain = None
+        if self.gateway_interface:
+            self.bridge_domain = self.gateway_interface.bridge_domain
 
     @property
     def _rest_definition(self):
```
mutax, line 69: I know you did not touch the bottom of that method, but I think adding a `return None` could improve readability so one does not overlook that there is an `elseif` and None can be a valid return value. (that should link to the property _rest_definition)

asr1k_neutron_l3/models/neutron/l3/prefix.py
sebageek: As this is a class variable we have under our control I'd expect that to always be either 4 or 6.

asr1k_neutron_l3/models/neutron/l3/nat.py
sebageek: I think I'd leave it like it is for now. We have two cases there and we don't have any explicit handling of `None` values. We could think about using enums or making sure specific values are not allowed here, but before doing that we're probably doing the `dict` refactoring. ;)

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: I think that has been done in a later commit.

mutax (dismissed):

sven-rosenzweig (dismissed):

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -434,3 +465,38 @@ def to_dict(self, context):
...
+        if self.prefix:
+            return self.prefix.split("/")[0]
+        return None
+
+
+class TrafficFilter(NyBase):
+    @classmethod
+    def __parameters__(cls):
+        return [
+            {"key": 'direction'},
```
mutax, line 494: Is it intentionally to use string literals here and not the constants defined? I wouldn't be surprised it there is a reason - if yes what is the criteria?

asr1k_neutron_l3/tests/common/fixtures.py
```diff
@@ -0,0 +1,231 @@
...
+                raise ValueError("Test cannot have ext_subnet and external_gateway_info")
+            external_gateway_info = {'network_id': ext_subnet['subnet']['network_id']}
+
+        # create router
+        ctx = context.get_admin_context()
+        with mock.patch.object(asr1k_db.DBPlugin, 'get_network_port_count_per_agent',
+                               return_value={'fake-agent': 0}):
+            with self.router(admin_state_up=admin_state_up, tenant_id=tenant_id,
+                             external_gateway_info=external_gateway_info, **kwargs) as router:
+                pass
```
mutax, line 183: adding a comment what this is doing / why this single pass is required would be highly appreciated by this reviewer but is not required. something like check if a router can be created without raising an exception

mutax (dismissed):

asr1k_neutron_l3/tests/common/fixtures.py
sebageek: This is a common pattern used in nearly all Neutron test code (common to Neutron, not common in general). The `with` is not really doing anything on entry/exit and I have no good idea why they use it this way expect that it might "look good". The original code can be found [here](https://github.com/sapcc/neutron/blob/266ae8ad642f37b72404655e7a70ded00fc3e2dc/neutron/tests/unit/extensions/test_l3.py#L469-L472).

As it is such a common pattern I'd leave it uncommented here.

sven-rosenzweig (dismissed):

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: There is no good reason for this.

The constants have always been there and are used in every `to_dict()` method, but they are nearly never used in the `__parameters__()` method. This is how I got the codebase and how I largely carried on the work. When writing new classes I in some cases used the constants, but while doing so I also noticed that in some cases this became very tedious: It works well for the `key` paramter, but it made the `yang-path` parameter completely unreadable. So, while the constants help against typos, it made it much harder to validate the `yang-path` against new firmware versions. Therefore I largely kept the way it was originally written and instead decided to introduce some tests for the classes.

I think it could make sense to replace the key argument with the already existing constants in some cases - wherever it doesn't hinder the readability, but we don't yet have anything policywise, so currently it's more or less a case-by-case decision.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -434,3 +465,38 @@ def to_dict(self, context):
...
+    LIST_KEY = L3Constants.PREFIX_LIST
+
+    @classmethod
+    def __parameters__(cls):
+        return [
+            {"key": 'prefix', 'id': True},
+        ]
+
+    def to_dict(self, context):
+        return {L3Constants.PREFIX: self.prefix.lower()}
```
mutax, line 481: using `self.ITEM_KEY` instead feels cleaner, also in case at some point somebody creates a subclass.

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -263,17 +283,18 @@ def postflight(self, context, method):
         LOG.debug("Processing Address Families")
         afs = []
         try:
-            afs = AddressFamily.get_for_vrf(context=context, asn=self.asn, vrf=self.id)
+            afs_v4 = AddressFamilyV4.get_for_vrf(context=context, asn=self.asn, vrf=self.id)
+            afs_v6 = AddressFamilyV6.get_for_vrf(context=context, asn=self.asn, vrf=self.id)
+            afs = afs_v4 + afs_v6
 
             if len(afs) == 0:
                 LOG.info("No address fammilies to clean")
```
mutax: could fix the typo in families

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -327,6 +348,14 @@ def to_dict(self, context):
         return dict(address_family)
 
 
+class IpV4AddressFamily(IpAddressFamilyBase):
```
mutax, line 350: hmm. We now have `IpV4AddressFamily` and `AddressFamilyV4` in this file, same for v6.
Is this too confusing on the long run or are we fine with that?
What is our opinion on importing AddressFamilyV4 as AddressFamilyV4_bgp or give the classes defined in this file a more distinctive name? `VrfIPv4AddressFamily` looks ugly, though

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -115,19 +114,13 @@ def _to_global_vrf_id(self, rt):
...
+    @property
+    def enable_ipv4(self):
+        return any(iface.ipv4_address for iface in self.interfaces.all_interfaces)
 
-    def get_routable_networks(self):
-        return [iface.primary_subnet['cidr'] for iface in self.address_scope_matches()
-                if iface.primary_subnet and 'cidr' in iface.primary_subnet]
+    @property
+    def enable_ipv6(self):
+        return any(iface.ipv6_addresses for iface in self.interfaces.all_interfaces)
```
mutax, line 123: will this trigger on fe80:: or ::1 on localhost? (not sure what is part of all_interfaces - does it query the router or neutron?)

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -222,40 +223,53 @@ def _build_nat_acl(self):
...
+                has_routable_interface = bool(self.interfaces.get_routable_networks(ip_version))
+                redistribute_map = f"bgp-redistribute{ip_version}-{utils.uuid_to_vrf_id(self.router_id)}"
+            else:
+                # use network statements
+                networks = self.interfaces.get_internal_networks(ip_version)
+                routable_networks = self.interfaces.get_routable_networks(ip_version)
+                has_routable_interface = bool(routable_networks)
+                redistribute_map = None
+
+            enable_af = self.enable_ipv4 if ip_version == 4 else self.enable_ipv6
```
mutax, line 250: (nitpick/more of a comment)
I know the chance that we will ever see something else than v4 or v6 in our lifetime are slim, but somehow not checking for ip_version==6 triggers my inner monk, also because it looks assymetrical to me as if something was missing 
`enable_af = self.enable_ipv4 and ip_version == 4 or self.enable_ipv6 and ip_version == 6`
would be an alternative and at least my brain processes it faster without thinking twice.

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -297,17 +311,77 @@ def _build_prefix_lists(self):
...
+        #   routable-extraroutes -> all extraroutes part of a dapnets
+        #   internal -> everything internal
+        #   extraroutes -> extraroutes that are not routable
+        for ip_version, PrefixClass in ((4, prefix.BasePrefixV4), (6, prefix.BasePrefixV6)):
+            if ip_version == 4 and not cfg.CONF.asr1k_l3.advertise_bgp_ipv4_routes_via_redistribute:
+                # we don't need the lists for v4 if they're not being used
+                # to avoid potential config locks, we won't configure them
+                continue
+
+            af_enabled = self.enable_ipv4 if ip_version == 4 else self.enable_ipv6
```
mutax, line 335: same as above

mutax (dismissed):

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: So far I've been trying to go with the constants in most cases and only use the `ITEM_KEY` if its needed by a subclass and is actually changing its value. I doubt we will ever subclass this one and I find the `PREFIX` easier to read... But to be honest I could go either way here, both options feel acceptable to me. Do you feel a benefit for the `ITEM_KEY` solution?

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: This information comes from Neutron, meaning whatever the user has specified. So it says "has the user requested anything that looks like v6?" --> we enable it.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: We have many places in which we make this exact distinction and I think I'd like to keep it this way. I understand your Inner Monk, but then I'd rather have us introduce an enum type for the AF and make sure we always have a 4 or 6 in there. Also, we don't have any handling for if this is not 4 or 6 in the way you propose, so my preference would be to leave it as it is for now.

asr1k_neutron_l3/models/netconf_yang/vrf.py
sebageek: Let's import just the module and not import the bgp classes into our module namespace. Makes it more readable :)

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
mutax: The only benefit I would see is making clear that the key in the dictionary is the same as the definition in the class - but I have no strong opinion. did a quick check, other classes don't use the `ITEM_KEY` as well. lets keep it that way, then its consistent in the file

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -233,16 +251,18 @@ def postflight(self, context, method):
...
             for route in routes:
-                LOG.info("Deleting hanging route {} in vrf {} postflight.".format(route.name, self.name))
+                LOG.info("Deleting hanging route %s in vrf %s postflight.", route.name, self.name)
                 route._delete(context=context)
-                LOG.info("Deleted hanging route {} in vrf {} postflight.".format(route.name, self.name))
-        except BaseException as e:
-            LOG.error("Failed to delete {} routes in VRF {} postlight : {}".format(len(routes), self.id, e))
+                LOG.debug("Deleted hanging route %s in vrf %s postflight.", route.name, self.name)
+        except Exception as e:
+            LOG.error("Failed to delete %s routes in VRF %s postlight: %s", len(routes), self.id, e)
```
mutax: typo light -> flight

also, as we are catching all exceptions here, do we want to log it as LOG.exception, so we get the stacktrace in the log as well?

sebageek: Corrected the typo. I presume we don't need the stack trace here, as it will most likely fail due to a device error, from which we only need the message to know what's up. Error reporting in the driver could be refactored, mid-term, though.

mutax (dismissed):

asr1k_neutron_l3/models/neutron/l3/vrf.py
```diff
@@ -34,18 +34,30 @@ def __init__(self, name, description=None, asn=None, rd=None, routable_interface
         self.asn = asn
         self.rd = utils.to_rd(self.asn, rd)
 
-        self.enable_bgp = False
-        self.map = "exp-{}".format(self.name)
+        # FIXME: hmmm might also need to be AF aware, but... when do we have a routable interface?
```
toanju:
```suggestion
        # NOTE: hmmm might also need to be AF aware, but... when do we have a routable interface?
```

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -263,17 +283,18 @@ def postflight(self, context, method):
...
+                LOG.info("No address families to clean")
 
             for af in afs:
-                LOG.info("Deleting hanging address family in vrf {} postflight.".format(self.name))
-                LOG.info(af)
+                LOG.info("Deleting hanging address family %s in vrf %s postflight.", af, self.name)
                 result = af._delete(context=context)
                 LOG.debug(result)
-                LOG.info("Deleted hanging address family in vrf {} postflight.".format(self.name))
+                LOG.debug("Deleted hanging address family %s in vrf %s postflight.", af, self.name)
```
toanju: could be merged with the previous LOG.debug

asr1k_neutron_l3/models/neutron/l3/bgp.py
```diff
@@ -21,17 +21,22 @@
...
-        self.routable_interface = routable_interface
+        self.enable_af = enable_af
+        self.has_routable_interface = has_routable_interface
         self.asn = asn
         self.enable_bgp = False
         self.rt_export = rt_export
+        self.redistribute_map = redistribute_map
         self.routable_networks = routable_networks
-        self.networks_v4 = set()
+        self.networks = set()
```
toanju, line 39: any reason why this is not a list from the start?

asr1k_neutron_l3/models/neutron/l3/interface.py
```diff
@@ -51,49 +52,115 @@ def all_interfaces(self):
 
         return result + self.internal_interfaces + self.orphaned_interfaces
 
+    def get_internal_networks(self, ip_version):
+        cidrs = []
+        for interface in self.internal_interfaces:
+            for subnet in interface.subnets:
+                if subnet.get('cidr') and utils.get_ip_version(subnet['cidr']) == ip_version:
+                    cidrs.append(subnet['cidr'])
+        cidrs.sort()
```
toanju, line 61: out of curiosity, do we need to have this sorted for any API or is this for the readability of a human being?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -405,8 +405,29 @@ def _update_router_gw_info(self, context, router_id, info,
...
         orig_router_atts = self.db.get_router_att(context, router_id)
         dynamic_nat_pool = None
-        if len(ext_ips) > 1:
-            dynamic_nat_pool = self._handle_dynamic_nat_pool_allocation(context, router_id, info, ext_ips)
+        ext_ips_v4 = []
+        ext_ips_v6 = []
+
+        for ext_ip in ext_ips:
+            ip_version = None
+            if 'subnet_id' in ext_ip:
```
sven-rosenzweig: Why do we need to fetch the ip version from the subnet? Can't we rely on the ip_address itself?

From my understanding, the info dict looks like: 
` "external_fixed_ips": [{"subnet_id": "uuid", "ip_address": "<ip-address>"}]`

sven-rosenzweig: Why not 
```
if 'ip_address' in ext_ip:
    blabla
 elif 'subnet_id' in ext_ip:
   blabla
```

asr1k_neutron_l3/models/neutron/l3/interface.py
sebageek: I sort these lists, as order is not guaranteed by OpenStack and is in fact considered to be unordered, but when this is the source for a prefix list or interface configuration, order might suddenly matter. Most classes do a sort themselves, but just on the off-chance that we missed it somewhere I like to keep it sorted.

In the past we had some prefix lists that changed their order on each update, causing a lot of config changes for "nothing has changed".

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: The `info` dict we're receiving here is not from the OpenStack database, but from the API request of the user, e.g. if the user types in `os router set $router_id --external-gateway $extnet --fixed-ip ip-address=$ip --fixed-ip subnet=$subnet` - this will result in an external gateway interface with two ips, one directly specified, one allocated from the subnet. we will only be having one of those entries in there in nearly all of the cases (I'm not sure adding both would be possible, but it would still be redundant). I can flip the if statement though, to prefer the IP if we happen to have both though.

asr1k_neutron_l3/models/neutron/l3/bgp.py
sebageek: Deduplication / distrust of the input parameters we get. Just to make sure there are no duplicate networks in there, which would result in yang trying to reconfigure the same object again and again, as "my lists looks different than what's on the device".

asr1k_neutron_l3/models/neutron/l3/vrf.py
sebageek: Uhm yeah, the comment in the code is just plain wrong. `self.map` is only used for v4 and v6 has its own map. I think that comment was added before this was the case (when I was still trying to figure out if I can do this with one route-map, which we could, but decided against it, due to other participants regarding this as a more complex solution).

I'll remove the offending comment completely.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sven-rosenzweig: thanks for the detailed explanation. Not need to flip this from my side

sven-rosenzweig (approved):

mutax (approved):
lgtm

asr1k_neutron_l3/models/neutron/l3/interface.py
toanju: whee thanks for the details

PR #114, opened 2024-08-14
Use route-maps instead of network for BGP prefixes

sebageek:
The current firmware of our Cisco routers (<= 17.6) requires a full sync every time we touch the bgp subtree via netconf-yang. For our current prod regions can cause a complete config lock of the /native tree, which can take up to two minutes. In this timeframe no other reconfiguration is possible.

Before this commit we used network statements for advertising BGPVPNs and DAPNets. This is now done via "redistribute connected" and "redistribute static" + a route-map, to set the right communities for extraroutes overlapping with a DAPNet. The config will still get locked for DAPNet routers or when a BGPVPN is attached, but all subsequent modifications (like adding/deleting a subnet or adding/deleting an extraroute to the router) will only change a prefixlist and therefore not touch the BGP config tree.

Long term it would be desirable if Cisco would fix their confd implementation, though.

WIP:
 * config is still hardcoded, shouldn't be that way
 * we have to discuss if we want to advertise DAPNets via route-map or network statement


sebageek:
Most of this functionality was implemented as part of #127, ipv4 can optionally be enabled to also go via the route maps.

PR #115, opened 2024-09-20
17.15 compatibility

sebageek:


asr1k_neutron_l3/models/netconf_yang/prefix.py
```diff
@@ -73,22 +77,37 @@ def __parameters__(cls):
...
+        for prefix in prefixes:
+            pfx_name = prefix.get(PrefixConstants.NAME)
+            if name is None:
+                name = pfx_name
+            elif name != pfx_name:
+                LOG.error("Could not fetch prefix-list - found two prefix-lists in yang: %s and %s - full data: %s",
+                          name, pfx_name, dict)
+                return
+
+            seqs.append(prefix)
```
swagner-de: Does this mean `remove_wrapper` supports only one prefix-list name per response? If I understood it correctly, all prefix lists are now flat in the `prefix-lists` key, meaning, if this gets called from the cleaner via a `cls.get_all_stubs_from_device` then this method will only report one prefix-list?

asr1k_neutron_l3/models/netconf_yang/prefix.py
```diff
@@ -107,43 +126,50 @@ def add_seq(self, seq):
...
+        dev_pfx_list = self._internal_get(context=context)
+        if dev_pfx_list and dev_pfx_list.seq:
+            neutron_seqs = {seq.no for seq in self.seq}
+            dev_seq_to_remove = []
+            for dev_seq in list(dev_pfx_list.seq):
+                if dev_seq.no in neutron_seqs:
+                    continue
+                LOG.warning("Prefix-list %s needs cleaning - seq %s is present on device but not in neutron",
+                            self.name, dev_seq)
+                dev_seq_to_remove.append(dev_seq)
```
swagner-de: Do you like set operations? :)
```
            dev_seq_to_remove = {seq.no for seq in dev_pfx_list.seq} - {seq.no for seq in self.seq}
            LOG.warning("Prefix-list %s needs cleaning - seqs %s are present on device but not in neutron",
                        self.name, dev_seq_to_remove)
```

asr1k_neutron_l3/models/netconf_yang/prefix.py
```diff
@@ -107,43 +126,50 @@ def add_seq(self, seq):
...
+
+            if dev_seq_to_remove:
+                dev_pfx_list.seq = dev_seq_to_remove
+                dev_pfx_list._delete(context=context)
```
swagner-de, line 148: This will cause a hickup in BGP depending on who is faster, Cisco's BGP implementation or our update, correct?

asr1k_neutron_l3/models/netconf_yang/prefix.py
sebageek: Yes, this is a problem. With the current code, no stubs will be found, as we will run into the "found two prefix-lists in yang" error line in the remove_wrapper method.

asr1k_neutron_l3/models/netconf_yang/prefix.py
sebageek: I remember either you or Ben doing some internal tests on this for the "let's announce our bgpvpn prefixes via prefix-lists"-task and the result was something like it takes 200ms to 5 seconds for Cisco's BGP implementation to pick up the change, so we should have more than enough time for this.

swagner-de (approved):

sven-rosenzweig (approved):

PR #116, opened 2024-10-11
Register DEFAULT group on agent start

sven-rosenzweig:
From upstream commit 4d3a2747 fixing bug (#1968606), registration of the group DEFAULT does not happen automatically anymore.

If not loaded, agent fails with 'oslo_config.cfg.NoSuchOptError: no such option api_extensions_path in group [DEFAULT]'

### DIFF
--- a/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
+++ b/asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
@@ -88,6 +88,7 @@ urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 def main(manager='asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgentWithStateReport'):
     asr1k_config.register_common_opts()
     asr1k_config.register_l3_opts()
+    common_config.register_common_config_options()
     common_config.init(sys.argv[1:])
     common_config.setup_logging()
     # set periodic interval to 10 seconds, as I understand the code this means
--- a/asr1k_neutron_l3/plugins/ml2/agents/asr1k_ml2_agent.py
+++ b/asr1k_neutron_l3/plugins/ml2/agents/asr1k_ml2_agent.py
@@ -65,6 +65,7 @@ SYNC_ROUTERS_MIN_CHUNK_SIZE = 32
 
 
 def main():
+    common_config.register_common_config_options()
     common_config.init(sys.argv[1:])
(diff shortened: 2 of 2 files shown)


sebageek (approved):

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
```diff
@@ -88,6 +88,7 @@
 def main(manager='asr1k_neutron_l3.plugins.l3.agents.asr1k_l3_agent.L3ASRAgentWithStateReport'):
     asr1k_config.register_common_opts()
     asr1k_config.register_l3_opts()
+    common_config.register_core_common_config_opts()
```
sebageek: Are you sure `common_config` is the right module? We import it as `from neutron.common import config as common_config` but the method is defined in `neutron.conf.common` (for stable/yoga).

sebageek (changes requested):
Is the import correct?

asr1k_neutron_l3/plugins/l3/agents/asr1k_l3_agent.py
sven-rosenzweig: Yes, import is right (for stable/2024.1). (https://github.com/sapcc/neutron/blob/0aa154b5ce9dc8da73309fb212843a2b69b68696/neutron/common/config.py#L52). Fixing the [bug](https://bugs.launchpad.net/neutron/+bug/1968606) makes this necessary.

sebageek: The commit you're referencing is not part of `stable/yoga-m3`. I can't find the method in the referenced file in `stable/yoga-m3`, meaning if we're going to merge it we're going to break the asr driver on yoga release - oooor I looked wrong and it's imported from somewhere? But if I'm right we need to either fix the import for yoga or we only fix this in Caracal, meaning you'd need to open this PR against `stable/2024.1-m3`.

sven-rosenzweig: Thx for pointing me to this. I wanted to open the PR against `stable/2024.1-m3`.

sebageek: As per our call let's use `register_common_config_options()` from `neutron.common.config`.

sebageek (approved):

swagner-de (approved):

PR #117, opened 2024-10-14
Drop all compatibility for pre 17.6 fw versions

sebageek:
We are now on version >= 17.6 everywhere and therefore we can drop all code for previous versions (16.9, 17.3). This also makes writing new code easier, as we have to consider (and test) less versions (and even don't have any devices anymore to test the old code).

~~For now I left the BDI code intact, as my main goal was to get rid of all version_min_17_x and drop_on_17_x, but we can also think about removing that in the future.~~ BDI code has been removed in second commit.

### DIFF
--- a/asr1k_neutron_l3/models/asr1k_pair.py
+++ b/asr1k_neutron_l3/models/asr1k_pair.py
@@ -29,7 +29,7 @@ LOG = logging.getLogger(__name__)
 class ASR1KContextBase(object):
     @property
     def use_bdvif(self):
-        return self.version_min_17_3
+        return True
 
     @property
     def bd_iftype(self):
@@ -44,10 +44,8 @@ class FakeASR1KContext(ASR1KContextBase):
     stable results
     """
 
-    def __init__(self, version_min_17_3=True, version_min_17_6=True, version_min_17_13=True, version_min_17_15=True,
+    def __init__(self, version_min_17_13=True, version_min_17_15=True,
                  has_stateless_nat=True):
-        self.version_min_17_3 = version_min_17_3
-        self.version_min_17_6 = version_min_17_6
         self.version_min_17_13 = version_min_17_13
         self.version_min_17_15 = version_min_17_15
         self._has_stateless_nat = has_stateless_nat
@@ -58,8 +56,6 @@ class FakeASR1KContext(ASR1KContextBase):
 
 
 class ASR1KContext(ASR1KContextBase):
-    version_min_17_3 = property(lambda self: self._get_version_attr('_version_min_17_3'))
-    version_min_17_6 = property(lambda self: self._get_version_attr('_version_min_17_6'))
     version_min_17_13 = property(lambda self: self._get_version_attr('_version_min_17_13'))
     version_min_17_15 = property(lambda self: self._get_version_attr('_version_min_17_15'))
     has_stateless_nat = property(lambda self: self._get_version_attr('_has_stateless_nat'))
@@ -108,8 +104,6 @@ class ASR1KContext(ASR1KContextBase):
 
             ver = tuple(_to_int_if_possible(d) for d in ver)
 
-            self._version_min_17_3 = ver >= (17, 3)
-            self._version_min_17_6 = ver >= (17, 6)
             self._version_min_17_13 = ver >= (17, 13)
             self._version_min_17_15 = ver >= (17, 15)
             self._has_stateless_nat = ver >= (17, 4)
@@ -137,7 +131,7 @@ class ASR1KContext(ASR1KContextBase):
 
     @property
     def use_bdvif(self):
-        return self.version_min_17_3 and self._use_bdvif and not self.force_bdi
+        return self._use_bdvif and not self.force_bdi
 
     def mark_alive(self, alive):
         if not alive:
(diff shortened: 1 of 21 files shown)


sven-rosenzweig (approved):

asr1k_neutron_l3/models/netconf_yang/bgp.py
```diff
@@ -82,14 +82,6 @@ def __parameters__(cls):
...
             {'key': 'asn', 'id': True, 'yang-key': 'id'},
             {'key': 'vrf', 'yang-key': 'name'},
-            {'key': 'connected', 'yang-path': 'ipv4-unicast/redistribute', 'default': False,
-             'yang-type': YANG_TYPE.EMPTY},
-            {'key': 'static', 'yang-path': 'ipv4-unicast/redistribute', 'default': False,
-             'yang-type': YANG_TYPE.EMPTY},
-            {'key': 'connected', 'yang-path': 'ipv4-unicast/redistribute-vrf', 'default': False,
-             'yang-type': YANG_TYPE.EMPTY},
-            {'key': 'static', 'yang-path': 'ipv4-unicast/redistribute-vrf', 'default': False,
-             'yang-type': YANG_TYPE.EMPTY},
```
ikatza-at-work, line 92: Just to clarify: why can we remove this code from the model? It's not clear to me from the commit message

ikatza-at-work (approved):

asr1k_neutron_l3/models/netconf_yang/bgp.py
sebageek: in the `to_dict()` method of the `AddressFamily` class we have a block using connected/static, which was only used on versions older than 17.3. With newer firmware we announce networks via bgp network statement directly and not via `redistrbute connected/static`. See comment `# redistribute connected/static is only used with 16.9`.

I planned to reintroduce this to circumvent some fw locking in #114 but will reintroduce it once we've decided to go with this route.

asr1k_neutron_l3/models/asr1k_pair.py
```diff
@@ -27,13 +27,7 @@
 
 
 class ASR1KContextBase(object):
```
swagner-de, line 29: Couldn't we just remove that class and let the other 2 not inherit from it? Otherwise I suppose you wanna have the joint base class to match the Test-Context and the real thing together.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -71,65 +70,49 @@ class L3Constants(object):
...
+                        </BD-VIF>
                     </interface>
                 </native>
              """
 
-    VRF_XPATH_FILTER = "/native/interface/{iftype}/vrf[forwarding='{vrf}']"
+    VRF_XPATH_FILTER = "/native/interface/BD-VIF/vrf[forwarding='{vrf}']"
 
     LIST_KEY = L3Constants.INTERFACE
+    ITEM_KEY = L3Constants.BDVIF_INTERFACE
```
swagner-de, line 112: Help me out, why do we need `ITEM_KEY` now?

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -232,10 +200,8 @@ def to_dict(self, context):
             ip[L3Constants.ADDRESS][L3Constants.PRIMARY][L3Constants.ADDRESS] = self.ip_address.address
             ip[L3Constants.ADDRESS][L3Constants.PRIMARY][L3Constants.MASK] = self.ip_address.mask
 
-        if self.nat_stick or (self.nat_inside and context.version_min_17_3):
+        if self.nat_stick or self.nat_inside:
```
swagner-de, line 203: We never set `nat_stick` on the object, we set it implicitly in the yang data model for the internal interfaces. If we would be correct, we'd set it instead of `nat_inside` and ditch that one completely. I see, except in `InternalInterface`  it's only ever used in `BDInterface.init_config()` which itself we would only be called by some RPC endpoint we don't use.

No need to change, just a comment and sth for some time later.

swagner-de (approved):

asr1k_neutron_l3/models/asr1k_pair.py
sebageek: Yeah, we could just remove that class. I left it there, as I "just removed the code we don't need" and thought a common base class wouldn't be to bad either way, maybe we want to refactor some of the common functionality into that one.

I can remove it if you'd like, otherwise I'd just leave it for now.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: There are two ways go get an `ITEM_KEY`. Either via the attribute or via the classmethod `get_item_key()`. The latter was introduced for models needing to return different keys depending on the asr1k context. As this is no longer dynamic we remove `get_item_key()` and use `ITEM_KEY` with a fixed value instead.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: We could remove `nat_inside` from the neutron `InternalInterface` class and specify `nat_stick` instead. But removing `nat_inside` altogether would probably the way to go then and with that we'd probably also would want to remove the whole `init_config()` code paths. So... maybe something for a followup?

PR #122, opened 2024-12-02
Subnet scope db and api and fwaas gw protection

sebageek:


asr1k_neutron_l3/extensions/asr1koperations.py
```diff
@@ -196,6 +199,16 @@ def update(self, request, id, **kwargs):
...
 
+    def sync_data(self, request, id, **kwargs):
+        check_access(request)
+        try:
+            data = self.plugin.get_sync_data(request.context, router_ids=[id])
+            if not data:
+                raise exceptions.HTTPNotFound(detail="No router found for given uuid")
+            return {"config": data[0]}
+        except BaseException as e:
+            raise exceptions.HTTPInternalServerError(detail=str(e))
```
nikatza, line 212: Is there a chance to leak some authetication credential? (contained in "e")

asr1k_neutron_l3/extensions/asr1koperations.py
```diff
@@ -196,6 +199,16 @@ def update(self, request, id, **kwargs):
         except BaseException as e:
             raise exceptions.HTTPInternalServerError(detail=str(e))
 
+    def sync_data(self, request, id, **kwargs):
+        check_access(request)
+        try:
+            data = self.plugin.get_sync_data(request.context, router_ids=[id])
+            if not data:
+                raise exceptions.HTTPNotFound(detail="No router found for given uuid")
```
nikatza: May be add the UUID for if we want to inspect lock messages

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -649,6 +649,17 @@ def set_router_dynamic_nat_pool(self, context, router_id, dynamic_nat_pool):
             ra = self.get_router_att(context, router_id)
             ra.dynamic_nat_pool = dynamic_nat_pool
 
+    @db_api.CONTEXT_READER
+    def get_subnet_address_scope_map(self, context, subnet_ids):
+        query = context.session.query(models_v2.Subnet.id,
+                                      models_v2.SubnetPool.address_scope_id)
+        query = query.outerjoin(
```
nikatza, line 656: why "outer" but not "inner"? (what case do we want to get from an outer join that we don't get with the inner?

nikatza (dismissed):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -300,6 +300,14 @@ def get_sync_data(self, context, router_ids=None, active=None, host=None):
                     if gw_info is not None:
                         gw_info['external_fixed_ips'] = gw_port['fixed_ips']
 
+            # add address scope to internal subnets
+            all_ports = router['_interfaces'] + ([gw_port] if gw_port else [])
+            all_subnets = {sn['id'] for port in all_ports for sn in port.get('subnets', [])}
+            subnet_address_scope_map = self.db.get_subnet_address_scope_map(context, all_subnets)
+            for port in all_ports:
+                for subnet in port.get('subnets', []):
```
mutax, line 308: you could make your code twice as fast using () instead of [] :)
The empty tuple is basically for free because it already exists and because it is immutable you always will get the same instance, while an empty list takes some time to instantiate. Not relevant here, but I find it good to know.
```
$ python -m timeit "x=()"
50000000 loops, best of 5: 7.08 nsec per loop
$ python -m timeit "x=[]"
20000000 loops, best of 5: 14.9 nsec per loop
```

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -300,6 +300,14 @@ def get_sync_data(self, context, router_ids=None, active=None, host=None):
...
+                    subnet['address_scope_id'] = subnet_address_scope_map.get(subnet['id'])
```
mutax, line 309: Could it become a problem if the subnet is not present in the map and we set the key to None?
Or would guarding it with a check if its None make sense to not set the address_scope_id key in the subnet dictionary at all? I don't know if the key is mandatory, so we actually need to set it to something and I know my grep-foo isn't that good, looking through the neutron code - just asking to ensure this was considered.

asr1k_neutron_l3/extensions/asr1koperations.py
mutax: came here to say the same, hence the thumbs up, I am not the author ^^ (nikatza marked the conversation as resolved after i thumbed it up)

mutax (dismissed):
just nitpics

asr1k_neutron_l3/extensions/asr1koperations.py
sebageek: It's the response that the user gets when asking for an ID, so they should know it... but can do it for the logs, yeah. Makes sense.

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: We do an outer join so we get a result for each subnet, even if it has no subnet pool attached, as in these cases we want to have `None` as address scope, i.e. subnet exists in dict, address scope `None`.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Nice idea, I'll keep it in mind. As we do it with the list in all other cases in the driver I'll keep it with the list for now.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: The address scope can just be `None`, that is actually not a problem. This just means that either the subnet has no subnetpool or no address scope, which is the case for all networks that are internal and are not DAPNets or IPv6 networks.

mutax (dismissed):

asr1k_neutron_l3/plugins/db/asr1k_db.py
nikatza: Ok, so `None` has also a meaning that we use, that is what i needed to know (also answered in another comment). Thanks!

mutax (approved):
lgtm

sven-rosenzweig (approved):

PR #126, opened 2025-02-21
Clean up PoolDynamicNat on orphan VRF delete

sebageek:
When a router is deleted, we clean up all objects relating to a VRF before deleting the VRF itself on the device. If this fails, we have the device cleaner cleaning up the orphaned objects, but for this to work we have to also make sure depending objects are deleted. This is already done for InterfaceDynamicNat, but not for PoolDynamicNat - so we now do it for PoolDynamicNat as well.

### DIFF
--- a/asr1k_neutron_l3/models/netconf_yang/vrf.py
+++ b/asr1k_neutron_l3/models/netconf_yang/vrf.py
@@ -24,7 +24,7 @@ from asr1k_neutron_l3.common import cli_snippets
 from asr1k_neutron_l3.common import utils
 from asr1k_neutron_l3.models.connection import ConnectionManager
 from asr1k_neutron_l3.models.netconf_yang.l3_interface import BDInterface
-from asr1k_neutron_l3.models.netconf_yang.nat import InterfaceDynamicNat
+from asr1k_neutron_l3.models.netconf_yang.nat import InterfaceDynamicNat, PoolDynamicNat
 from asr1k_neutron_l3.models.netconf_yang.ny_base import NyBase, Requeable, NC_OPERATION, execute_on_pair, \
     retry_on_failure, YANG_TYPE
 from asr1k_neutron_l3.models.netconf_yang.route import VrfRouteV4, VrfRouteV6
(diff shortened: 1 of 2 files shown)


asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -234,17 +234,22 @@ def postflight(self, context, method):
...
-                      "".format(len(interface_nats), self.id, e))
+            if len(all_vrf_nats) > 0:
+                for vrf_nat in all_vrf_nats:
+                    LOG.info("Deleting hanging interface NAT %s %s in VRF %s postflight",
+                             vrf_nat.__class__.__name__, vrf_nat.id, self.name)
+                    vrf_nat._delete(context=context)
+                    LOG.info("Deleted hanging interface NAT %s %s in VRF %s postflight",
+                             vrf_nat.__class__.__name__, vrf_nat.id, self.name)
+            else:
+                LOG.debug("No interface NAT to clean for %s", self.id)
```
mutax: would change the log message to indicate that no interface and no pool nats have been found

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -234,17 +234,22 @@ def postflight(self, context, method):
...
-            for interface_nat in interface_nats:
-                LOG.info("Deleting hanging interface nat {} in vrf {} postflight.".format(interface_nat.id, self.name))
-                interface_nat._delete(context=context)
-                LOG.info("Deleted hanging interface nat {} in vrf {} postflight.".format(interface_nat.id, self.name))
-        except BaseException as e:
-            LOG.error("Failed to delete {} interface NAT in VRF {} postlight : {}"
-                      "".format(len(interface_nats), self.id, e))
+            if len(all_vrf_nats) > 0:
+                for vrf_nat in all_vrf_nats:
+                    LOG.info("Deleting hanging interface NAT %s %s in VRF %s postflight",
```
mutax: is there a need to indicate it it is a pool or interface nat? if no, would it make sense to reword?

asr1k_neutron_l3/models/netconf_yang/vrf.py
```diff
@@ -234,17 +234,22 @@ def postflight(self, context, method):
...
+                             vrf_nat.__class__.__name__, vrf_nat.id, self.name)
+                    vrf_nat._delete(context=context)
+                    LOG.info("Deleted hanging interface NAT %s %s in VRF %s postflight",
+                             vrf_nat.__class__.__name__, vrf_nat.id, self.name)
+            else:
+                LOG.debug("No interface NAT to clean for %s", self.id)
+        except Exception as e:
+            LOG.error("Failed to delete %s interface NAT in VRF %s postlight: %s",
```
mutax: this is referencing interface_nats in len, not all_vrf_nats - intentionally?

mutax (approved):

felix-kaestner (approved):

PR #132, opened 2025-03-12
Better dynamic nat pool gateway allocation logic

sebageek:
For dynamic NAT pooling we always have one block of consecutive IPv4
addresses and one gateway address, which does not need to be part of the
block. If the gateway IP is not specified by the user we left it to
Neutron to allocate a fitting IP, which randomly allocates one. This
caused problems for users as this random IP might fragment up the IP
space of their network, resulting in less routers with NAT pools fitting
in the network.

To fix this we now use a more deterministic approach: We allocate the
highest available IP in the network. As we always try to use the
smallest block with the lowest starting IP for the NAT pool we should
end up having less fragmentations in the IP space. Other methods we
thought of were using first available after pool or first available
around pool but this might interfere with other pool allocations in case
we have many routers entering and leaving the network (it also might
not, I don't have any reliable data on this - the decision is based on
educated guesses).

We also now retry the allocation in the rare case that in between
selecting and allocating the IP someone else allocated it, with up to 5
tries and a random wait between 0 and 0.05s. The retry is only done when
an IpAddressAlreadyAllocated exception occurs. This might also lead to a
retry if the user specifies already allocated options, leading to a few
more DB queries and a slower query of 0.2s.

For retrying we use tenacity, which is also used by neutron. The asr1k
driver already uses retrying, but this library is no longer maintained
and we should switch to tenacity for everything else in the foreseeable
future.

The allocation logic for the NAT pool itself now also changes slightly,
in that we make sure that we sort all available IP address space chunks
not only by size, but also by their first IP. So in cases where we have
multiple chunks of the same size we should end up with the chunk with
the smallest starting IP.

### DIFF
--- a/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
+++ b/asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
@@ -14,6 +14,7 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 from collections import OrderedDict
+import copy
 from operator import attrgetter
 import time
 
@@ -23,6 +24,7 @@ from neutron.db import extraroute_db
 from neutron.db import l3_gwmode_db as l3_db
 from neutron.extensions.tagging import TAG_PLUGIN_TYPE
 from neutron.ipam import driver as neutron_ipam_driver
+from neutron.ipam import exceptions as ipam_exc
 from neutron_lib.api.definitions import availability_zone as az_def
 from neutron_lib.api.definitions import l3 as l3_def
 from neutron_lib.callbacks import events
@@ -34,6 +36,7 @@ from neutron_lib.plugins import directory
 from oslo_config import cfg
 from oslo_log import helpers as log_helpers
 from oslo_log import log
+import tenacity
 
 from asr1k_neutron_l3.common import asr1k_constants as constants
 from asr1k_neutron_l3.common import asr1k_exceptions as asr1k_exc
@@ -400,8 +403,14 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
         result = super(ASR1KPluginBase, self).create_router(context, router)
         return result
 
+    @tenacity.retry(retry=tenacity.retry_if_exception_type(ipam_exc.IpAddressAlreadyAllocated),
+                    stop=tenacity.stop_after_attempt(5), reraise=True,
+                    wait=tenacity.wait_random(min=0.0, max=0.05))
     def _update_router_gw_info(self, context, router_id, info,
                                request_body, router=None):
+        # we are modifying the info dict, but need the original in case of a retry --> deepcopy()
+        info = copy.deepcopy(info)
+
         ext_ips = info.get('external_fixed_ips', []) if info else []
         orig_router_atts = self.db.get_router_att(context, router_id)
         dynamic_nat_pool = None
@@ -474,7 +483,6 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
 
             for ext_ip, ip in zip(ext_ips, ips):
                 ext_ip['ip_address'] = ip
-            return dynamic_nat_pool
         else:
             # all ips are specified, check that they are consecutive
             ipset = netaddr.IPSet([ip_def['ip_address'] for ip_def in ext_ips[:-1]])
@@ -492,26 +500,46 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
                     break
             else:
                 raise asr1k_exc.DynamicNatPoolGivenIPsDontBelongToNetwork(ip=pool_from, network_id=info['network_id'])
+
+        # try to allocate a gateway ip that doesn't fragment up the space too much
+        gw_ip_def = ext_ips[-1]
+        if 'subnet_id' in gw_ip_def and 'ip_address' not in gw_ip_def:
+            nat_pool_ips = [netaddr.IPAddress(ext_ip['ip_address']) for ext_ip in ext_ips[:-1]]
+            ip = self._find_gateway_ip_for_dynamic_nat_pool(context, gw_ip_def['subnet_id'], nat_pool_ips)
+            if ip:
+                gw_ip_def['ip_address'] = str(ip)
+            else:
+                LOG.warning("Failed to find gateway ip in subnet %s for router %s with nat pool %s, "
+                            "falling back to letting OpenStack find one (is the IP space exhausted?)",
+                            gw_ip_def['subnet_id'], router_id, dynamic_nat_pool)
+
         return dynamic_nat_pool
+
+    def _get_subnet_allocation_pools_and_allocations(self, context, subnet_id):
+        """Get a subnet's allocation pools and an IPSet of already allocated IPs for a subnet"""
+        ipam_driver = neutron_ipam_driver.Pool.get_instance(None, context)
+        ipam_subnet = ipam_driver.get_subnet(subnet_id)
+        allocation_pools = ipam_subnet.subnet_manager.list_pools(context)
+        allocations = ipam_subnet.subnet_manager.list_allocations(context)
+        ip_allocations = netaddr.IPSet([netaddr.IPAddress(allocation.ip_address) for allocation in allocations])
+
+        return allocation_pools, ip_allocations
 
     def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
         """Search for a block of consecutive IPs in the given subnet, return ips and pool"""
-        ipam_driver = neutron_ipam_driver.Pool.get_instance(None, context)
-        ipam_subnet = ipam_driver.get_subnet(subnet_id)
-        allocations = ipam_subnet.subnet_manager.list_allocations(context)
-        ip_allocations = netaddr.IPSet([netaddr.IPAddress(allocation.ip_address) for allocation in allocations])
+        allocation_pools, ip_allocations = self._get_subnet_allocation_pools_and_allocations(context, subnet_id)
 
         plugin = directory.get_plugin()
         subnet = plugin.get_subnet(context, subnet_id)
 
-        for ip_pool in ipam_subnet.subnet_manager.list_pools(context):
+        for ip_pool in allocation_pools:
             ip_set = netaddr.IPSet()
             ip_set.add(netaddr.IPRange(ip_pool.first_ip, ip_pool.last_ip))
             av_set = ip_set.difference(ip_allocations)
             if av_set.size < ip_count:
                 continue
 
-            av_ranges = sorted(av_set.iter_ipranges(), key=attrgetter('size'))
+            av_ranges = sorted(av_set.iter_ipranges(), key=attrgetter('size', 'first'))
             for av_range in av_ranges:
                 if av_range.size >= ip_count:
                     result = [str(ip) for ip in av_range[:ip_count]]
@@ -519,6 +547,24 @@ class ASR1KPluginBase(l3_db.L3_NAT_db_mixin,
                     return result, pool
 
         return None, None
+
+    def _find_gateway_ip_for_dynamic_nat_pool(self, context, subnet_id, nat_pool_ips):
+        """Find a usable gateway ip for a nat pool in a subnet without fragmenting the ip space too much"""
+        # current allocation logic: last ip that is not allocated and not in the pool
+        allocation_pools, ip_allocations = self._get_subnet_allocation_pools_and_allocations(context, subnet_id)
+
+        gw_ip_candidate = None
+        for ip_pool in allocation_pools:
+            for ip in netaddr.iter_iprange(ip_pool.last_ip, ip_pool.first_ip, step=-1):
+                # don't need to look at smaller ips if we have a candidate
+                if gw_ip_candidate and ip < gw_ip_candidate:
+                    break
+
+                if ip not in ip_allocations and ip not in nat_pool_ips:
+                    gw_ip_candidate = ip
+                    break
+
+        return gw_ip_candidate
 
     def ensure_default_route_skip_monitoring(self, context, router_id, router):
         tag_plugin = directory.get_plugin(TAG_PLUGIN_TYPE)
--- a/asr1k_neutron_l3/tests/unit/plugins/l3/service_plugins/test_l3_extension_adapter.py
+++ b/asr1k_neutron_l3/tests/unit/plugins/l3/service_plugins/test_l3_extension_adapter.py
@@ -65,12 +65,12 @@ class TestASR1kExtensionAdapter(test_l3.L3BaseForIntTests, test_l3.L3NatTestCase
                 router_atts = db.get_router_att(ctx, router['router']['id'])
                 self.assertEqual("10.100.1.2-10.100.1.5/24", router_atts.dynamic_nat_pool)
 
-                expected_pool_ips = [f"10.100.1.{n}" for n in range(2, 6)]
+                expected_pool_ips = [f"10.100.1.{n}" for n in range(2, 6)] + ["10.100.1.254"]
                 router_ips = [ip_def['ip_address']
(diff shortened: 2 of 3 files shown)


asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -400,8 +403,14 @@ def create_router(self, context, router):
         result = super(ASR1KPluginBase, self).create_router(context, router)
         return result
 
+    @tenacity.retry(retry=tenacity.retry_if_exception_type(ipam_exc.IpAddressAlreadyAllocated),
+                    stop=tenacity.stop_after_attempt(5), reraise=True,
+                    wait=tenacity.wait_random(min=0.0, max=0.05))
     def _update_router_gw_info(self, context, router_id, info,
                                request_body, router=None):
+        # we are modifying the info dict, but need the original in case of a retry --> deecopy()
```
nikatza: typo: deepcopy()

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -492,26 +500,41 @@ def _handle_dynamic_nat_pool_allocation(self, context, router_id, info, ext_ips)
                     break
             else:
                 raise asr1k_exc.DynamicNatPoolGivenIPsDontBelongToNetwork(ip=pool_from, network_id=info['network_id'])
+
+        # try to allocate a gateway ip that doesn't fragment up the space too much
+        gw_ip_def = ext_ips[-1]
```
nikatza, line 505: we are not checking that `ext_ips` is empty, right? Can't this raise an exception?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -520,6 +543,24 @@ def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
...
+                # don't need to look at smaller ips if we have a candidate
+                if gw_ip_candidate and ip < gw_ip_candidate:
+                    break
+
+                if ip not in ip_allocations and ip not in nat_pool_ips:
+                    gw_ip_candidate = ip
+
+        if not gw_ip_candidate:
+            return None
+        return str(gw_ip_candidate)
```
nikatza: why do we need to return a string representation of the ip? May be that is another function's problem :)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -492,26 +500,41 @@ def _handle_dynamic_nat_pool_allocation(self, context, router_id, info, ext_ips)
...
         ipam_driver = neutron_ipam_driver.Pool.get_instance(None, context)
         ipam_subnet = ipam_driver.get_subnet(subnet_id)
+        allocation_pools = ipam_subnet.subnet_manager.list_pools(context)
         allocations = ipam_subnet.subnet_manager.list_allocations(context)
         ip_allocations = netaddr.IPSet([netaddr.IPAddress(allocation.ip_address) for allocation in allocations])
 
+        return allocation_pools, ip_allocations
+
+    def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
+        """Search for a block of consecutive IPs in the given subnet, return ips and pool"""
```
nikatza, line 529: Suggestion: "Search for the smallest block of at least N (ip_count) consecutive IPs in a given subnet...."

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -492,26 +500,41 @@ def _handle_dynamic_nat_pool_allocation(self, context, router_id, info, ext_ips)
...
-    def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
-        """Search for a block of consecutive IPs in the given subnet, return ips and pool"""
+    def _get_subnet_allocation_pools_and_allocations(self, context, subnet_id):
         ipam_driver = neutron_ipam_driver.Pool.get_instance(None, context)
         ipam_subnet = ipam_driver.get_subnet(subnet_id)
+        allocation_pools = ipam_subnet.subnet_manager.list_pools(context)
         allocations = ipam_subnet.subnet_manager.list_allocations(context)
         ip_allocations = netaddr.IPSet([netaddr.IPAddress(allocation.ip_address) for allocation in allocations])
 
+        return allocation_pools, ip_allocations
```
nikatza: I was confused here. May be moving this function after `_find_ips_for_dynamic_nat_pool` would help with the flow, otherwise may be add a docstring, or even copy these few lines into `_find_ips_for_dynamic_nat_pool`if this is the only call being done. (I don't like large functions but with my limited knowledge I can't straightly follow relying on the naming)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
nikatza: What is the difference between a set of ips and a pool?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -520,6 +543,24 @@ def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
 
         return None, None
 
+    def _find_gateway_ip_for_dynamic_nat_pool(self, context, subnet_id, nat_pool_ips):
+        # current allocation logic: last ip that is not allocated and not in the pool
+        allocation_pools, ip_allocations = self._get_subnet_allocation_pools_and_allocations(context, subnet_id)
+
+        gw_ip_candidate = None
+        for ip_pool in allocation_pools:
```
nikatza, line 557: I leave a comment here for me to comment late (since i can't follow without more info)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: `_handle_dynamic_nat_pool_allocation()` is called by `_update_router_gw_info()`. It's only called if we have at least two external ipv4 addresses, so it should always have at least two entries.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Hmyeah you're right, the caller can take care of this.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: We return a list of ips that is part of the pool (for the code to later use in... anything) and a pool definition, which is needed for the db, containing the ip block and the prefix length of the subnet the ips are in. Does that help?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
nikatza: I see. I miss c++ assertions :)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
nikatza: Wondering if we could pack the return objects  together in a object (not saying that it has to be done, just wondering if it would make sense)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -520,6 +548,23 @@ def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
 
         return None, None
 
+    def _find_gateway_ip_for_dynamic_nat_pool(self, context, subnet_id, nat_pool_ips):
+        """Find a usable gateway ip for a nat pool in a subnet without fragmenting the ip space too much"""
+        # current allocation logic: last ip that is not allocated and not in the pool
+        allocation_pools, ip_allocations = self._get_subnet_allocation_pools_and_allocations(context, subnet_id)
+
+        gw_ip_candidate = None
+        for ip_pool in allocation_pools:
```
nikatza, line 557: we are iterating twice over allocations_pools (essentially a subset of the iteration in line 535). Would it make sense to change this?

nikatza (approved):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Python has the `assert` statement, we could in theory do something like `assert len(ext_ips) > 0`, though this has not been "the style" of the code base so far. Could be an option, though.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
nikatza: my error, i confused the names. closing :)

nikatza (dismissed):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: I think I would leave it as it is. I think creating a custom object wouldn't increase readability, but on the other hand maybe I've been in python too long as in that returning tuples feels so common. ;)

nikatza: i don't know the rest of the codebase so fine for me!

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
nikatza: The assert helps understand the assumptions about the input. I guess we remove assertions during runtime anyway so it wouldnt be detrimental, but i leave it this to you.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -520,6 +548,23 @@ def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
...
+            for ip in netaddr.iter_iprange(ip_pool.last_ip, ip_pool.first_ip, step=-1):
+                # don't need to look at smaller ips if we have a candidate
+                if gw_ip_candidate and ip < gw_ip_candidate:
+                    break
+
+                if ip not in ip_allocations and ip not in nat_pool_ips:
+                    gw_ip_candidate = ip
```
felix-kaestner, line 564:
```suggestion
                    gw_ip_candidate = ip
                    break
```
nit: I think we could also continue with the next allocation_pool here, since we are iterating in reverse order over all IPs, so we know that we would break in the next iteration on line 561 either way.

sebageek: That's true. I thought what would be nicer and opted for "just one `break`" but the other option felt equally nice to me at time of writing. As I like both, what do the reviewers vote? ;)

felix-kaestner: No strong opinions (hence also the "nit") so I leave that decision to you. ^^

felix-kaestner (dismissed):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -520,6 +548,23 @@ def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
 
         return None, None
 
+    def _find_gateway_ip_for_dynamic_nat_pool(self, context, subnet_id, nat_pool_ips):
```
mutax, line 551: if this method qould return more than one candidate, e.g. up to 10, the caller could try them in order, so we do not run the whole logic again. It also would prevent the situation that if the selected IP is broken for some reason and cannot be used, the next call would return the same IP again. With  a list the caller would just use the next one that then probably will work.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -400,8 +403,14 @@ def create_router(self, context, router):
         result = super(ASR1KPluginBase, self).create_router(context, router)
         return result
 
+    @tenacity.retry(retry=tenacity.retry_if_exception_type(ipam_exc.IpAddressAlreadyAllocated),
+                    stop=tenacity.stop_after_attempt(5), reraise=True,
+                    wait=tenacity.wait_random(min=0.0, max=0.05))
```
mutax, line 408: I would start with 0.01 or something - not a huge difference but prevents very fast retries. (nit, just feels nicer to the systems)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -492,26 +500,46 @@ def _handle_dynamic_nat_pool_allocation(self, context, router_id, info, ext_ips)
...
+            else:
+                LOG.warning("Failed to find gateway ip in subnet %s for router %s with nat pool %s, "
+                            "falling back to letting OpenStack find one (is the IP space exhausted?)",
+                            gw_ip_def['subnet_id'], router_id, dynamic_nat_pool)
+
         return dynamic_nat_pool
 
-    def _find_ips_for_dynamic_nat_pool(self, context, subnet_id, ip_count):
-        """Search for a block of consecutive IPs in the given subnet, return ips and pool"""
+    def _get_subnet_allocation_pools_and_allocations(self, context, subnet_id):
```
mutax, line 518: would it make sense to limit to Ipv4? or is this already happening? Could make the search faster if we only look at subnets that actually have v4?

mutax (dismissed):

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: Before this method is called we separate the request in v4 and v6 ips, as we only do nat pooling for v4, so we should only do this for v4.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: I would trust the randomness here. I think even without random wait we'd be fine, so I would leave it as is.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
mutax: In general I prefer to break/return/terminate as early as possible. When reading code, especially my own, I would probably be a bit confused and try to find out why the search was continued although it apparently already found a candidate. Maybe there is a side effect, maybe its needed, but why? what am I missing? 
also when later code is added, and be it for logging, one wonders why its executed after a hit

sebageek: We might have multiple allocation pools (we nearly never have that but still have to handle the case) and the second pool might be larger than the first. I'll add the extra break then. :)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: I think the added complexity is not worth the effort for how often we'll retry. If router creations with nat pools ever fail I will keep this in mind as an enhancement

mutax (approved):
lgtm

nikatza (approved):

felix-kaestner (approved):

PR #142, opened 2025-07-30
Fix: Unprocessed ports deleted from `updated_ports` and `deleted_ports`

swagner-de:
When a port update or creation comes in, the ml2 agent's `port_update`
method puts it into self.updated_ports.

From there the port_id is picked up by `process_ports` which gets a copy
of the dictionary, finds out which ports actually belong to this agent
and then runs netconf to configure them. At the end of its operation it
clears the global `self.updated_port`.

Ports that arrived while this was happening would be cleared alltogether
and hence have to wait until the next network syncloop picks them up.

Let us only delete the ports from the dictionary that we have actually
had in our copy.

The same problem also applies for the `deleted_ports`.


sebageek (approved):

sebageek:
Could add something about the `deleted_ports` as well into the commit message.

sebageek (approved):

felix-kaestner (approved):

PR #143, opened 2025-08-18
VPNaaS support for asr1k driver

sebageek:
Implement the Neutron VPNaaS API backend to support IPSec on Cisco XE
devices. This is a large commit, as most of the moving parts needed to
implement this only make sense together.

Architecture: A VPNaaS router in our infrastructure is not allowed to
have any subnets. Subnets need to be connected via BGPVPN. As we don't
know the local subnets, the user needs to specify them all via an
endpoint group. To make sure routers are scheduled on our VPNaaS
hardware, the vpnaas trait is required for them (and also required for
attaching a vpn service object to a router).

VPNaaS driver: We have a VPNaas service driver, much like our FWaaS and
BGPVPN service driver. It takes care of most of the validation via
registry hooks. In some places this replaces the direct calls used
normally by VPNaaS, as validation needs to happen inside the DB
transaction or needs to fetch some related data before some object is
deleted. In most cases we then notify the agent to do a router sync,
except for ipsec site connection deletion (see section on deletion
below). VPNaaS information is fetched as part of the get_sync_data() and
is then incorporated into the models.neutron.l3.Router class.

API extensions: For our VPNs we need configurable internal tunnel ips
(IPv4 and IPv6) and the ability to configure a NAT IP. Both these things
need to be configured by the user, so we provide API extensions for
these two cases. Each has their own api_def, which extends the
ipsec-site-connection resource.

Internal tunnel IPs: These IPs, provided by an extra extension, are used
for each ipsec site connection that we have. If they are not provided by
the user they are auto allocated from the link local pool of the address
family. We do some validation to make sure the IPs of site connections
on the same router do not overlap. The user can specify a local subnet
and a peer address. If the peer address is not part of the local subnet,
then a static route is added for this.

Tunnel ids: A Tunnel interface on Cisco hardware needs a number. As we
need to remember this number, we have a table in our db for it. The
number gets allocated from a configurable range on a per-agent basis and
is deleted once the ipsec site connection is deleted.

Neutron VPNaaS: We depend on our own brand of VPNaaS[0], as we need some
extra notifications (DB precommit/postcommit hooks) and resource
extender for extra API fields. Additionally, the "disconnected subnet
mode" config is needed for router-subnet validation.

Endpoint groups: We only allow cidr endpoint groups, as our VPNaaS
routers don't have directly connected subnets. We impose a maximum of
100 endpoints (by default, value is configurable) per endpoint group, as
in transport mode we need to create an ACL with source x destination
entries and decided 10000 is a good upper limit for ACL rule count.

Device/YANG: We require Cisco IOS XE >= 17.15, else some YANG objects
are not compatible. We configure: IPSecProfile, IPSecTransformSet,
IKEv2Profile, IKEv2Keyring, IKEv2Policy, IKEv2Proposal and
TunnelInterfaces. TunnelInterfaces bring up some YANG problems again, as
deleting a TunnelInterface causes a full sync of the device. This
might cause some problems when deleting / de-scheduling router and also
with the device cleaner.

Device cleaner: Some objects have shared custody (are both referenced
via vrf and a site connection / tunnel id). Objects related to a Tunnel
interface cannot be deleted as long as the Tunnel interface is up, so it
is one of the first objects that we need to delete.

IPSec site connection deletion: We don't clear all related objects when
a site connection is deleted, because a router object only
syncs/overwrites its current state and when a site connection is
deleted there is no longer state for this object. To make sure the
customer tunnel goes down, on site connection deletion we trigger a
Tunnel interface deletion on the device. To delete the Tunnel interface
we also need to delete all objects that depend on it, which currently
are only static routes which have this interface as a dependency. For
all other site connection objects on the device we trust on the device
cleaner to remove them in the next cleaning cycle. They also don't block
us and do not keep interesting state, so we can just leave them there
until the next cleaning.

Router / vpn service / ipsec site connection status: VPN Service and
ipsec site connection have a "status" field, which is set to
PENDING_CREATE by default. Our router only has ACTIVE/ERROR as state,
but this is updated whenever the router is synced by a regular update
(not by an API router sync). For VPNaaS we piggyback onto this mechanism
and set all related VPNaaS objects to ACTIVE when a router has been
synced successfully. For the admin-triggered API sync we also now set
the router status.

IPv6: Support for IPv6 is partly done. It has not been tested and there
are some open questions regarding how we map the OpenStack data model to
Cisco config when we cannot fully offer dual stack. For example, tunnel
mode can only be either ipv4 or ipv6 when the underlay is ipv6, while
with ipv4 underlay a dual stack overlay is allowed (by Cisco config).
This has not been fully mapped out and will be handled in a follow up.

[0] https://github.com/sapcc/neutron-vpnaas/tree/stable/2024.1-m3

---
 * Tests will fail until this PR is merged: https://github.com/sapcc/neutron-vpnaas/pull/4
 * IPv6 might not be fully functional yet
 * configuration is present on device, but no data path tests have been done yet


asr1k_neutron_l3/db/migration/alembic_migrations/versions/2024.1/expand/e0739ed2c92c_add_vpnaas_tunnel_ip_and_id_tracking.py
```diff
@@ -0,0 +1,65 @@
...
+def upgrade():
+    op.create_table(
+        'asr1k_tunnel_ids',
+        sa.Column('agent_host', sa.String(length=36), nullable=False),
+        sa.Column('ipsec_site_connection_id',
+                  sa.String(length=36),
+                  sa.ForeignKey('ipsec_site_connections.id', ondelete='CASCADE'),
+                  nullable=False, primary_key=True),
+        sa.Column('number', sa.Integer(), nullable=False),
+        sa.UniqueConstraint('agent_host', 'number', name='agent_host0number'),
```
mutax: TODO(mutax): host0number? why the 0?

asr1k_neutron_l3/db/migration/alembic_migrations/versions/2024.1/expand/e0739ed2c92c_add_vpnaas_tunnel_ip_and_id_tracking.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    op.create_table(
+        'asr1k_internal_tunnel_ips',
+            sa.Column('ipsec_site_connection_id',
+                      sa.String(length=36),
+                      sa.ForeignKey('ipsec_site_connections.id', ondelete='CASCADE'),
+                      nullable=False, primary_key=True),
+            sa.Column('local_cidr_v4', sa.String(64), nullable=False),
```
mutax, line 51: Is there a specific reason to use 64 chars here? prefixes and addresses aren't that long. Just a safeguard and for being consistent?

asr1k_neutron_l3/extensions/vpn_asr1k_int_tunnel_ips.py
```diff
@@ -0,0 +1,94 @@
...
+    API_PREFIX = '/vpn'
+
+    RESOURCE_ATTRIBUTE_MAP = {
+        vpn.IPSEC_SITE_CONNECTIONS: {
+            # internal cidr v4
+            'int_local_cidr_v4': {
+                'allow_post': True, 'allow_put': True,
+                'default': constants.ATTR_NOT_SPECIFIED,
+                'convert_to': converters.convert_cidr_to_canonical_format,
+                'validate': {'type:subnet_or_none': None},
```
mutax, line 40: do we need to make sure that only addresses/subnets with the correct IP version are supplied? Could something break later / when configuring?

asr1k_neutron_l3/extensions/vpn_asr1k_nat_address.py
```diff
@@ -0,0 +1,67 @@
...
+    API_PREFIX = '/vpn'
+
+    RESOURCE_ATTRIBUTE_MAP = {
+        vpn.IPSEC_SITE_CONNECTIONS: {
+            # internal cidr v4
+            'peer_nat_address_v4': {
+                'allow_post': True, 'allow_put': True,
+                'default': constants.ATTR_NOT_SPECIFIED,
+                'convert_to': converters.convert_ip_to_canonical_format,
+                'validate': {'type:ip_address_or_none': None},
```
mutax, line 40: same as earlier - ensure at validation that it's IPv4?

asr1k_neutron_l3/models/netconf_yang/crypto.py
```diff
@@ -0,0 +1,539 @@
...
+                }
+            }
+
+        if self.reverse_route:
+            profile[CryptoConstants.REVERSE_ROUTE] = ""
+
+        return {CryptoConstants.PROFILE: profile}
+
+    def is_orphan_vpnaas(self, all_ipsec_siteconnection_ids, all_tunnel_ids):
+        if not self.name or len(self.name) != 31:
```
mutax, line 157: maybe add a comment why 31 -- a uuid without would be 32 chars, so its not a uuid? or is this the device-firmware-limit?

asr1k_neutron_l3/models/netconf_yang/crypto.py
```diff
@@ -0,0 +1,539 @@
...
+            return utils.vrf_id_to_uuid(self.fvrf)
+        return None
+
+
+class IKEv2Proposal(NyBase):
+    ITEM_PATH = [CryptoConstants.CRYPTO]
+    LIST_KEY = CryptoConstants.IKEV2
+    LIST_KEY_NS = xml_utils.NS_CISCO_CRYPTO
+    ITEM_KEY = CryptoConstants.PROPOSAL
+    GROUPS = ['fourteen', 'fifteen', 'sixteen', 'nineteen', 'twenty', 'twenty_one']
```
mutax, line 497: I am guessing cisco is just doing cisco stuff here and seventeen and eighteen are not needed and .. twenty_one with underscore all is correct ...

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -500,3 +523,179 @@ def to_dict(self, context):
...
+        if self.tunnel_src:
+            tun[L3Constants.SOURCE] = self.tunnel_src
+        if self.tunnel_dest_ipv4 or self.tunnel_dest_ipv6:
+            dest_config = {}
+            if self.tunnel_dest_ipv4:
+                dest_config[L3Constants.IPV4] = self.tunnel_dest_ipv4
+            if self.tunnel_dest_ipv6:
+                dest_config[L3Constants.IPV6] = self.tunnel_dest_ipv6
+            tun[L3Constants.DESTINATION_CONFIG] = dest_config
+        if self.tunnel_mode_ipsec_ipv4:
```
mutax, line 658: Only one of these tunnel modes can be active at a time, I get that (the key is overwritten) but this also logs no error / raises an exception when for some reason the object is broken and the last one active wins. But I also get that there is no validation done at this point because that should have happened earlier already. But there is a chance that this hides an error? (nitpick)

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
```diff
@@ -500,3 +523,179 @@ def to_dict(self, context):
...
+
+    def postflight(self, context, method):
+        if self.name:
+            # clear all routes referencing this tunnel interface
+            for route_cls in (route.VrfRouteV4, route.VrfRouteV6):
+                route_cls.delete_routes_by_nexthop(context, f"Tunnel{self.name}")
+
+    def is_orphan_vpnaas(self, all_ipsec_siteconnection_ids, all_tunnel_ids):
+        if not self.name:
+            return
```
mutax: this can return True/False/None - intentionally?

asr1k_neutron_l3/models/netconf_yang/route.py
```diff
@@ -122,6 +125,28 @@ def to_delete_dict(self, context):
...
+        """Clear all routes referencing a specific nexthop"""
+        all_on_device = cls._get_all(context=context, filter={})
+        routes_found = False
+        for vrf_routes in all_on_device:
+            if not vrf_routes.neutron_router_id:
+                # just to be on the safe side we're ignoring everything not handled by the driver
+                # even if it references something we're trying to delete
+                continue
+
+            new_routes = [r for r in vrf_routes.routes if r.fwd_list and r.fwd_list.get('fwd') != nexthop]
```
mutax, line 139: Is this a string comparison and could the representation of  the nexthop change? e.g. adding a prefixlen/adding or removing leading 0es? Or do we need to convert to an ipaddress and compare that?
(if that is a user supplied value e.g. via API call, this could be necessary anyway to normalize  it?)

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -114,13 +116,23 @@ def _to_global_vrf_id(self, rt):
                       rt, global_vrf_id)
         return global_vrf_id
 
+    def get_status_set(self):
+        status = {self.status}
+        if self.router_info.get('vpn'):
+            status.add(self.router_info['vpn']['status'])
+            for sitecon in self.router_info['vpn']['ipsec_site_connections']:
```
mutax, line 123: could at some point there be more types of vpns, so we would have a router_info with a dictionary in vpn that has no ipsec_site_connections key so we would also guard that here with a get to precent a KeyError?

(Such a change will most likely require major rework anyway so that's ok the way it is, then?)

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -190,6 +202,24 @@ def _build_routes(self):
             if self._route_has_connected_interface(r):
                 routes[ip_net.version].append(r)
 
+        # handle vpn routes (endpoint group + int tun ip static route)
+        if self.router_info.get('vpn'):
```
mutax, line 206: in the enable_ipvX methods the comparison casts to bool explicitly, could make it consistent and either not use bool in that methods or use bool here as well.

asr1k_neutron_l3/models/neutron/l3/router.py
```diff
@@ -190,6 +202,24 @@ def _build_routes(self):
...
+            for sitecon in self.router_info['vpn']['ipsec_site_connections']:
+                target_dest = f"Tunnel{sitecon['tun_iface_id']}"
+                for ep_net in sitecon['peer_ep_group']['endpoints']:
+                    ep_net = netaddr.IPNetwork(ep_net)
+                    RouteClass = route.RouteV4 if ep_net.version == 4 else route.RouteV6
```
mutax, line 211: I think that I have seen that pattern several times in the code: checking the IP version and then using the respective class for v4/v6.  (TODO: check where before sending review. could have been this commit but yesterday)

I also found it here:
https://github.com/sapcc/asr1k-neutron-l3/blob/0cfd058a02f395c8e4f850f5e80659600a994636/asr1k_neutron_l3/models/neutron/l3/router.py#L187

Now, when looking at
https://github.com/sapcc/asr1k-neutron-l3/blob/0cfd058a02f395c8e4f850f5e80659600a994636/asr1k_neutron_l3/models/neutron/l3/route.py#L49

We could 
- introduce a RouteBaseClass and give that one a classmethod that works like a factory
- or have an extra helper method somewhere that inspects the ip version of the argument

Could make the code easier to read and prevent typos/errors

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import ipaddress
+
+from oslo_log import log as logging
+
+from asr1k_neutron_l3.common.utils import uuid_to_ipsec_short_id, uuid_to_vrf_id
+from asr1k_neutron_l3.models.neutron.l3 import base
```
mutax: imports that come from the same modules, can be moved to the same lines

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+        if sitecon['peer_id'] or sitecon['peer_nat_address_v4']:
+            identities_v4 = []
+            if sitecon['peer_id']:
+                identities_v4.append(crypto.IKEv2ProfileIdentityV4(address=sitecon['peer_id'],
+                                                                   mask="255.255.255.255"))
+            if sitecon['peer_nat_address_v4']:
+                identities_v4.append(crypto.IKEv2ProfileIdentityV4(address=sitecon['peer_nat_address_v4'],
+                                                                   mask="255.255.255.255"))
+            extra_args['remote_identities_v4'] = identities_v4
+        if lifetime := sitecon['ikepolicy']['lifetime']:
```
mutax, line 128: add a newline before the if lifetime to make the extra_args assignment above visually easier to spot?

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+        return f"ikev2_profile_{self._rest_definition.name}"
+
+
+class IPSecTransformSet(base.Base):
+    def __init__(self, vrf, ipsecpolicy):
+        super().__init__()
+
+        self.name = self.gen_name_from_policy(ipsecpolicy)
+
+
```
mutax: expecting one empty line here ^^

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+    @property
+    def id(self):
+        return self.name
+
+    @classmethod
+    def _get_crypto_args_from_policy(cls, ipsecpolicy):
+        args = {}
+        enc = ipsecpolicy['encryption_algorithm']
+        key_bit = enc.split("-")[1]
```
mutax: could assign the key_bit directly and skip introducing the extra variable here (although one can argue that its better to read to have it)

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+        self.shutdown = not (vpn_service_up and sitecon['admin_state_up'])
+
+        self.tunnel_id = sitecon['tun_iface_id']
+        local_cidr_v4 = ipaddress.ip_interface(sitecon['int_local_cidr_v4'])
+        local_cidr_v6 = ipaddress.ip_interface(sitecon['int_local_cidr_v6'])
+
+        extra_args = {}
+        if sitecon['ipsecpolicy']['encapsulation_mode'] == 'transport':
+            # NOTE(seba): unsetting this attribute might need a replace, but for now upstream does not allow
+            #             editting of ipsec policies that are in-use
```
mutax: typo in editting

asr1k_neutron_l3/models/neutron/l3/vpn.py
```diff
@@ -0,0 +1,306 @@
...
+            extra_args['tunnel_mode_ipsec_ipv4'] = True
+            extra_args['tunnel_src'] = external_v6_ip
+
+        self._rest_definition = l3_interface.TunnelInterface(
+            name=self.tunnel_id,
+            shutdown=self.shutdown,
+            vrf=vrf,
+            tunnel_vrf=vrf,
+            description=sitecon['id'],
+            mtu=sitecon['mtu'], tcp_mss=sitecon['mtu'] - 40,
```
mutax: I'd add a comment explaining the magic number here. 
We don't need a sanity check that the value in the dict is an int, and the resulting number is still positive?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
+# Copyright 2025 SAP SE
+#
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http:..www.apache.org.licenses.LICENSE-2.0
```
mutax: that link looks broken - was this file target of a search and replace operation / -> . ?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+from asr1k_neutron_l3.common.asr1k_exceptions import InvalidVPNaaSIPSecSiteConnectionConfig
+from asr1k_neutron_l3.common.asr1k_exceptions import MultipleVPNServicesOnRouterDisallowed
+from asr1k_neutron_l3.common.asr1k_exceptions import PeerNatAddressMustBeIPv4
+from asr1k_neutron_l3.common.asr1k_exceptions import RouterIsNotVPNaaSFlavor
+from asr1k_neutron_l3.common.cache_utils import get_cache
+
+
+LOG = logging.getLogger(__name__)
+
+# device support + BSI approved, aes-{128,256} map to cbc
```
mutax, line 52: maybe add the current date here or - if we have it handy  - a link to the TR document of BSI and the current version - they get updated regularly and this could make it easier in the future to check if the list here needs an update too. I was unable to find the specific one, not sure if it is BSI TR-02102-1

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+
+        tun_ids = self.l3db.get_vpn_tunnel_ids(context, ipsec_site_connection_ids=[sitecon_id])
+        cache.set(f"delete-{sitecon_id}-tun-info", tun_ids)
+
+    @registry.receives(v_const.IPSEC_SITE_CONNECTION, [events.AFTER_DELETE])
+    def _delete_ipsec_site_connection_send_tunnel_send_notify(self, resource, event, trigger, payload):
+        context = payload.context
+        sitecon_id = payload.resource_id
+        cache = get_cache()
+        if not cache:
```
mutax, line 119: When we do not get a cache, I see that the `get_cache()` method will log an error, that a cache is required to delete routers. In case we run into that issue this message could be a bit misleading, especially because we are not logging that we failed to delete the tunnel(interface) here because of that.
(same above in line 111)

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+        tun_ips = self._autoallocate_int_tun_ips(context, tun_ips, ipsec_sitecon_id, vpnservice_id)
+        self.l3db.create_or_update_vpn_internal_tunnel_ips(context, ipsec_sitecon_id,
+                                                           tun_ips['local_cidr_v4'], tun_ips['peer_address_v4'],
+                                                           tun_ips['local_cidr_v6'], tun_ips['peer_address_v6'])
+
+        self.validate_internal_tunnel_ips(context, vpnservice_id)
+
+    def _autoallocate_int_tun_ips(self, context, tun_ips, ipsec_sitecon_id, vpnservice_id):
+        all_vpn_tun_ips = None
+        for ip_ver, def_subnet, def_prefixlen in ((4, self.INT_IPV4_NET, 30),  (6, self.INT_IPV6_NET, 126)):
```
mutax, line 202: they don't like using a /31 for point to point, correct? could save us some IPv4's :)
update: ok, would make the allocation logic a bit more complex

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+                if net_a.ip == ip_b or net_b.ip == ip_a:
+                    reason = (f"IPSec Site Connection {ips_a['ipsec_site_connection_id']} and "
+                              f"{ips_b['ipsec_site_connection_id']} have overlapping local/peer IPs")
+                    raise InvalidVPNaaSInternalTunnelIps(reason=reason)
+
+
+            for net_key in ('local_cidr_v4', 'local_cidr_v6'):
+                if to_net(ips_a[net_key]).overlaps(to_net(ips_b[net_key])):
+                    reason = (f"Local subnet {ips_a[net_key]} of {ips_a['ipsec_site_connection_id']} "
+                              f"and {ips_b[net_key]} of {ips_b['ipsec_site_connection_id']} overlaps")
```
mutax: overlap not overlaps?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+                    raise InvalidVPNaaSInternalTunnelIps(reason=reason)
+
+            for ip_key in ('peer_address_v4', 'peer_address_v6'):
+                if ipaddress.ip_address(ips_a[ip_key]) == ipaddress.ip_address(ips_b[ip_key]):
+                    reason = (f"Address {ips_a[ip_key]} cannot be used by two peers, it is used by "
```
mutax: what is the difference to the error case above in line 277? The Error message is almost identical -- difference is using _Peer address_ vs. _Address_ -- this might be confusing when hunting an issue to distinguish the errors?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+        return asr1k_const.TRAIT_VPNAAS in (flavor_info["req_traits"] + flavor_info["opt_traits"])
+
+    def _ensure_value(self, value, supported_values, obj_name, key_name):
+        if value not in supported_values:
+            raise InvalidVPNaaSIPSecSiteConnectionConfig(obj=obj_name, key=key_name, value=value,
+                                                         supported_values=", ".join(supported_values))
+
+    @db_api.CONTEXT_READER
+    def validate_ipsec_site_connection(self, context, ipsec_sitecon):
+        LOG.info("validate_ipsec_site_connection() called for ipsec_sitecon: %s", ipsec_sitecon)
```
mutax: info -> debug?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+
+        # get and validate ikepolicy
+        if 'ikepolicy_id' in ipsec_sitecon:
+            ikepolicy = self.driver.service_plugin.get_ikepolicy(context, ipsec_sitecon['ikepolicy_id'])
+            self._ensure_value(ikepolicy['auth_algorithm'], DEVICE_SUPPORTED_AUTH_ALGOS, "IKE Policy", "auth algorithm")
```
mutax, line 347: add constants for these values that are used in several places here, "IKE Policy", "IPSec Policy", ... ?
It often feels like a typo waiting to happen somewhere (missing space, wrong capitalization)...

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+            router = self.l3_plugin.get_router(context, vpn['router_id'])
+            if not router:
+                return
+            if not self._is_vpnaas_flavor(context, router['flavor_id']):
+                raise RouterIsNotVPNaaSFlavor(router_id=vpn['router_id'])
+
+    @registry.receives(v_const.VPNSERVICE, [events.PRECOMMIT_CREATE])
+    def _vpnservice_disallow_setting_subnet_id(self, resource, event, trigger, payload):
+        vpn = payload.states[0]
+        if vpn['subnet_id']:
```
mutax, line 393: so we expect the key to be present in the dict but to be empty? would the field missing also be okay? then we should use get or catch the KeyError. I would expect a deprecated field to be removed at some point, if its globally deprecated.

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -49,6 +50,7 @@
 
 from asr1k_neutron_l3.common import asr1k_constants as constants
 from asr1k_neutron_l3.common import asr1k_exceptions
+from asr1k_neutron_l3.common import utils
```
mutax: could go on the same line as the previous one

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
```diff
@@ -112,7 +116,46 @@ def clean_fwaas(self, context, dry=False):
...
+            all_tunnel_ids = self.plugin_rpc.get_tunnel_ids(context)
+            LOG.debug("Got %d IPSec Site Connections and %d tunnel ids on device",
+                      len(all_ipsec_siteconnection_ids), len(all_tunnel_ids))
+
+            for device, objs in device_objects.items():
+                for obj in objs:
+                    if not obj.is_orphan_vpnaas(all_ipsec_siteconnection_ids=all_ipsec_siteconnection_ids,
+                                                all_tunnel_ids=all_tunnel_ids):
+                        continue
+                    LOG.debug("Cleaning %s %s on device %s", obj.id, obj.__class__.__name__, device.host)
```
mutax, line 144: same log message for dryrun?

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
```diff
@@ -112,7 +116,46 @@ def clean_fwaas(self, context, dry=False):
...
+                for device in ASR1KPair().contexts:
+                    objs = cls.get_all_stubs_from_device(device)
+                    LOG.debug(f"Got {len(objs)} {cls.__name__} from device {device.host}")
+                    device_objects[device].extend(objs)
+
+            LOG.debug("Fetching all VPNaaS data on this agent from neutron")
+            all_ipsec_siteconnection_ids = self.plugin_rpc.get_ipsec_site_connection_ids(context)
+            all_tunnel_ids = self.plugin_rpc.get_tunnel_ids(context)
+            LOG.debug("Got %d IPSec Site Connections and %d tunnel ids on device",
+                      len(all_ipsec_siteconnection_ids), len(all_tunnel_ids))
```
mutax, line 137: these should all be sets, right?

asr1k_neutron_l3/plugins/l3/rpc/rpc_api.py
```diff
@@ -19,6 +19,7 @@
 
 from asr1k_neutron_l3.common import cache_utils
 from asr1k_neutron_l3.common.instrument import instrument
+from asr1k_neutron_l3.common import utils
```
mutax: can go to the import above

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -204,11 +212,18 @@ def __init__(self):
...
 
         self.db = asr1k_db.get_db_plugin()
+        self._vpn_plugin = None
         self._refresh_flavor_quotas_loop = loopingcall.FixedIntervalLoopingCall(self._refresh_flavor_qutotas)
         if cfg.CONF.asr1k.flavor_quota_refresh_interval > 0:
             self._refresh_flavor_quotas_loop.start(interval=cfg.CONF.asr1k.flavor_quota_refresh_interval,
                                                    stop_on_exception=False, initial_delay=1)
 
+    @property
+    def vpn_plugin(self):
```
mutax, line 222: no check to `utils.is_vpnaas_enabled()` neccessary/wanted?

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -353,8 +368,45 @@ def get_sync_data(self, context, router_ids=None, active=None, host=None):
                 all_port_ids = [port["id"] for port in all_ports]
                 router["fwaas_policies"] = self.get_fwaas_policies(context, all_port_ids)
 
+            router["vpn"] = None
+            if utils.is_vpnaas_enabled() and router["flavor_id"]:
+                # only routers with an attached flavor can have VPNs
```
mutax, line 373: but we do not care about the specific flavor, right?

asr1k_neutron_l3/db/migration/alembic_migrations/versions/2024.1/expand/e0739ed2c92c_add_vpnaas_tunnel_ip_and_id_tracking.py
sebageek: I thought about using the exact length, making the v4 and v6 fields different insize, also depending on if it's a cidr or not. Then I saw that neutron in many places just uses 64 for an ipaddress so I went with this as well. So I'd say... consistency.

asr1k_neutron_l3/extensions/vpn_asr1k_int_tunnel_ips.py
sebageek: Yes, but I haven't seen neutron having an extra validator for that. We should handle this in `vpnaas_driver.py` in `validate_internal_tunnel_ips()`

asr1k_neutron_l3/models/netconf_yang/crypto.py
sebageek: This is a firmware limit. This specific object has a name limit of exactly 31 chars. This is also why `common.utils.uuid_to_ipsec_short_id()` shrinks the uuid to 31 chars.

asr1k_neutron_l3/models/netconf_yang/crypto.py
sebageek: It's not a Cisco thing, it's a NIST thing. This is the intersection of NIST-approved and Cisco-implemented ciphers. The rest are rejected by the API. `twenty_one` is the name of the attribute, this will be converted to `twenty-one` in YANG.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: The class should be used in a way that only one mode is active, but we always have two data sources. First data source is models.neutron.l3.router.Router / some related object, which is filled with OpenStack info. Due to the way the API works we always should only have one of the two values in here and it should be fine. The other source is the device. If the device shows me in its YANG output that both tunnel modes are there, I want to see this in the model, so that once we do a `router-diff` on it we'll see that there's something different on the device. Both should not be possible from YANG side, but well.... We've both seen our fair share of software :)

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
sebageek: Actually no good reason for that.

asr1k_neutron_l3/models/netconf_yang/route.py
sebageek: The data from the device should be normalized regarding IP addresses. In the way this method is used though this is a string comparsion, as the only type of nexthop we're deleting routes for are Tunnel interfaces, i.e. `"Tunnel1234"`.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: I think for that we'd need vpn flavor support and do a major rework of the way. `get_vpnaas_objects()` is called as part of `get_sync_data()` and should always ensure that the key is there if we have any vpn contents whatsoever.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: In this case we're just checking if something is there, the cast to bool is implicit, as always with python `if`s. This is different from the `enable_ipvx` methods, as those use an `or`, which expands to the value of the arguments (`a or b` gets expanded to `a if a else b`. In some places we use `self.enable_ivp4` as an argument to methods that MIGHT put that into a yang request and I want to make sure that we have a bool there and not suddenly all of the vpn definitions.

asr1k_neutron_l3/models/neutron/l3/vpn.py
sebageek: Ah, the lines _below_ this line! Got it, yeah.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: `TR-02102-3` was the one we've been looking into afaik. Will add it.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: I'll change the error message to something more generic.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: We can use `/31` on the connection and the user can also specify that. For auto allocation I went with /30, as I found it to be more common in my usage-examples, though I don't have the data anymore. With the link local space we should have enough IPs though, so I don't think this will become a problem.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: They're used as arguments for `InvalidVPNaaSIPSecSiteConnectionConfig()` and only used as human readable names in the error message, so I think it's fine.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: It's in the [API def](https://github.com/sapcc/neutron-lib/blob/4e4ae002da6ec806c799a6cfaeebc00564a576a0/neutron_lib/api/definitions/vpn.py#L246-L248) and should always be defined, defaults to `None`.

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
sebageek: Yeah, same way as with the FWaaS cleaner. We don't often do dry_runs (they're only triggered by the API). I think if we wanna refactor this we could do it in a followup, what do you think?

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
sebageek: They're lists. No sets involved. As they're coming directly out of the DB they shouldn't include duplicates (as a feature of the queries that we do, not because a DB is inherently immune to having duplicates)

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: The check is on the other side here. If some piece of code thinks of calling any function on the vpn_plugin they should first check if it's actually enabled. If it's not enabled we run either in a "None does not have this method" or an explicit exception. So... usage side.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sebageek: At the moment only routers which have a flavor_id are even eligible for having a VPNaaS router. As the number of routers that have a flavor attached (and for the foreseeable future will have a flavor attached) is pretty low I thought I could do a small optimization here and prevent us querying for VPNaaS objects. Querying the specific traits for the flavor would be extra queries and "if the VPN objects are already there let's just configure them". The specific flavor (the one with a "req_traits VPNaaS") should only be relevant when scheduling and when attaching the initial VPN Service object.

TL;DR yes, you're right, we don't care about the specific flavor :)

mutax: Aye, my though was that if we have at some point a high number of routers with flavors that are not relevant for vpnaas we would not need to query the DB for them -- but what you say makes sense, would be additional queries or some hardcoded values in the config or something and we do not need that optimization here yet.

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
mutax: I am usually in favor of marking dryruns so when looking at logs I can see if something was a dryrun - but it depends on how often we expect dryruns to happen. I am happy to declare this a non-issue :)

asr1k_neutron_l3/extensions/vpn_asr1k_int_tunnel_ips.py
mutax: We could also implement a validator and try to get that upstream later, it just feels like a lack of input validation here and does not follow the 'fail early' -- but it is what it is: an existing code base that we need to adapt to. Would add a comment in the source though that this does not validate the IP version and we are doing that in the other method to document that decision/behavior.

asr1k_neutron_l3/models/netconf_yang/crypto.py
mutax: I would appreciate a short comment whenever magic numbers occur - makes it easier to understand the code. I remembered that limit but was not sure if that applies here as well and when looking at this 31 always feels like an off-by-one.

asr1k_neutron_l3/models/netconf_yang/l3_interface.py
mutax: >  We've both seen our fair share of software :)

does that mean you are also in favor of logging an error or throwing an exception here when more than one is set?

asr1k_neutron_l3/models/neutron/l3/router.py
mutax: ah, yea, right, I missed the different behavior because I missed that its an `or` there :)

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+    def delete_ipsec_site_connection(self, context, ipsec_site_connection):
+        # this is handled via an AFTER_DELETE hook
+        pass
+
+    def update_ipsec_site_connection(self, context, old_ipsec_site_connection, ipsec_site_connection):
+        self._notify_router_for_ipsec_site_connection(context, ipsec_site_connection['vpnservice_id'])
+
+    def _notify_router_for_ipsec_site_connection(self, context, vpnservice_id):
+        vpn = self.service_plugin.get_vpnservice(context, vpnservice_id)
+        LOG.info("IPSec site connection of VPN service %s changed - updating router %s",
```
sven-rosenzweig, line 91: Does it make sense to indicate wether we `create` or `update` a site_connection in the log line? 
Leaving this up to you, just something that crossed my mind reading this line of code.

asr1k_neutron_l3/tests/common/fixtures.py
```diff
@@ -35,6 +40,11 @@
 from asr1k_neutron_l3.models.neutron.l3.router import Router
 from asr1k_neutron_l3.plugins.db import asr1k_db
 from asr1k_neutron_l3.plugins.l3.service_plugins.l3_extension_adapter import ASR1KPluginBase
+from asr1k_neutron_l3.tests.common.vpnaas import ASR1KVPNaaSMixin
+from asr1k_neutron_l3.tests.common.flavor import FlavorSchedulingMixin
+
+
+DB_VPN_PLUGIN_KLASS = "neutron_vpnaas.services.vpn.asr1k_plugin.VPNASR1KDriverPlugin"
```
mutax, line 47: Typo?

asr1k_neutron_l3/tests/common/fixtures.py
```diff
@@ -64,25 +75,39 @@ def get_request_extensions(self):
...
-        service_plugins = {'l3_plugin_name': l3_plugin}
+        service_plugins = {'l3_plugin_name': l3_plugin, 'vpnaas': DB_VPN_PLUGIN_KLASS}
+
+        self.node_driver = "asr1k_neutron_l3.neutron.services.service_providers.asr1k_router.ASR1KRouterDriver"
+        vpn_driver = 'asr1k_neutron_l3.neutron.services.service_drivers.asr1k.vpnaas_driver.ASR1KIPSecVPNaaSDriver'
+        cfg.CONF.set_override('service_provider', [f'L3_ROUTER_NAT:asr1k:{self.node_driver}:default',
+                                                   f'VPN:cisco_ipsec:{vpn_driver}:default'],
+                              group='service_providers')
+        cfg.CONF.set_override("router_scheduler_driver",
+                              "asr1k_neutron_l3.plugins.l3.schedulers.simple_asr1k_scheduler.SimpleASR1KScheduler")
```
mutax, line 92: With this the test cases always have the vpn driver and flavor support somewhat present in the fixture, right?
Do we still cover the case where a deployment would come e.g. without the vpnaas driver "installed" with our tests? Could that be(come) an issue?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+        self.l3db = asr1k_db.get_db_plugin()
+        super().__init__(service_plugin, ASR1KIpsecVpnValidator(self))
+
+    def create_rpc_conn(self):
+        self.agent_rpc = ASR1KIPSecVPNaaSNotifier()
+
+    def create_vpnservice(self, context, vpnservice_dict):
+        LOG.info("asr1k driver creating vpnservice: %s", vpnservice_dict)
+        return super().create_vpnservice(context, vpnservice_dict)
+
```
sven-rosenzweig:
```
def update_vpnservice(self, context, old_vpnservice, vpnservice):
   LOG.info("asr1k driver updating vpnservice: .....)
    return super().update_vpnservice(context, old_vpnservice, vpnservice)
```

For consistency reasons, does it help to override the `update_vpnservice` and `delete_vpnservice` methods as well?

asr1k_neutron_l3/tests/common/flavor.py
```diff
@@ -0,0 +1,81 @@
...
+
+        return self._make_agent(ctx, host, configuration=c)
+
+    def _make_meta(self, req=None, opt=None, quota=None):
+        metainfo = {}
+        if req is not None:
+            metainfo["req_traits"] = req
+        if opt is not None:
+            metainfo["opt_traits"] = opt
+
```
mutax, line 76: why this empty line here and none above the other if statements?

asr1k_neutron_l3/tests/common/vpnaas.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2013 Hewlett-Packard Development Company, L.P.
```
mutax, line 1: is there still HP/Cisco code left in this file?

asr1k_neutron_l3/tests/common/vpnaas.py
```diff
@@ -0,0 +1,229 @@
...
+                                      },
+                                      'vpnservice_id': vpnservice_id,
+                                      'ikepolicy_id': ikepolicy_id,
+                                      'ipsecpolicy_id': ipsecpolicy_id,
+                                      'admin_state_up': admin_state_up,
+                                      'tenant_id': self._tenant_id,
+                                      'local_ep_group_id': local_ep_group_id,
+                                      'peer_ep_group_id': peer_ep_group_id}
+        }
+        if kwargs.get('description') is not None:
```
mutax, line 93: add description to the arguments with None as default? 
That's the only use of kwargs and we might get rid of it then, helping IDEs to complete the argument list.

asr1k_neutron_l3/tests/common/vpnaas.py
```diff
@@ -0,0 +1,229 @@
...
+        data = {'ipsecpolicy': {'name': name,
+                                'auth_algorithm': auth_algorithm,
+                                'encryption_algorithm': encryption_algorithm,
+                                'encapsulation_mode': encapsulation_mode,
+                                'transform_protocol': transform_protocol,
+                                'lifetime': {'units': lifetime_units,
+                                             'value': lifetime_value},
+                                'pfs': pfs,
+                                'tenant_id': self._tenant_id}}
+        if kwargs.get('description') is not None:
```
mutax, line 179: same as above, adding description with a default of None?

asr1k_neutron_l3/tests/common/vpnaas.py
```diff
@@ -0,0 +1,229 @@
...
+                               name='endpointgroup1',
+                               endpoints=[],
+                               expected_res_status=None,
+                               as_admin=False,
+                               **kwargs):
+
+        data = {
+            'endpoint_group': {
+                'name': name,
+                'type': kwargs.get('type'),
```
mutax, line 199: add type=None to the declaration should do the same?

asr1k_neutron_l3/tests/unit/models/netconf_yang/test_crypto.py
```diff
@@ -0,0 +1,526 @@
...
+        self.assertEqual("3", prof.dpd_retry)
+        self.assertEqual("on-demand", prof.dpd_query)
+        self.assertEqual("secret-stash", prof.keyring_local)
+        self.assertEqual("4242", prof.lifetime_sec)
+        self.assertEqual("mediterranean-gull", prof.fvrf)
+        self.assertEqual(2, len(prof.remote_identities_v4))
+        self.assertEqual("23.23.23.23", prof.remote_identities_v4[0].address)
+        self.assertEqual("255.255.255.255", prof.remote_identities_v4[0].mask)
+        self.assertEqual("42.42.42.42", prof.remote_identities_v4[1].address)
+        self.assertEqual("255.255.255.255", prof.remote_identities_v4[1].mask)
```
mutax: use different values for  the masks in the testdata to ensure that they do not get mixed up in the from_xml method?

asr1k_neutron_l3/tests/unit/services/service_drivers/asr1k/test_vpnaas_driver.py
```diff
@@ -0,0 +1,316 @@
...
+        super().setUp()
+        cfg.CONF.set_override('disconnected_subnets_mode', True, group='vpnaas')
+        self.register_address_scope_rt("the-open-sea", "65123:101")
+        self.vpn_driver = directory.get_plugin('VPN').drivers['cisco_ipsec']
+
+        ctx = context.get_admin_context()
+        self.vpnaas_flavor_id = self._make_flavor(ctx, "vpnaas-falaaaaavor",
+                                                  profiles=[{'metainfo': self._make_meta(req=["vpnaas"])}])
+
+
```
mutax: pep (2 blank lines)

asr1k_neutron_l3/tests/common/fixtures.py
sebageek: Copypaste from the neutron-vpnaas codebase. Neutron has it as well. I think this is an evolutionary remnant of `class` being a reserved keyword, so for `@classmethod`, people use one of `cls`, `clazz` or `klass` as the first argument. Then someone went and needed to reference this, added `KLASS` to their variable and... here we are.

```
quoth:~/repos/neutron-vpnaas/neutron_vpnaas/tests$ grep KLASS -rI
unit/db/vpn/test_vpn_ext_gw_db.py:OVN_VPN_PLUGIN_KLASS = "neutron_vpnaas.services.vpn.ovn_plugin.VPNOVNPlugin"
unit/db/vpn/test_vpn_ext_gw_db.py:    def setUp(self, core_plugin=None, vpnaas_plugin=OVN_VPN_PLUGIN_KLASS,
unit/db/vpn/test_vpn_db.py:DB_CORE_PLUGIN_KLASS = 'neutron.db.db_base_plugin_v2.NeutronDbPluginV2'
unit/db/vpn/test_vpn_db.py:DB_VPN_PLUGIN_KLASS = "neutron_vpnaas.services.vpn.plugin.VPNPlugin"
unit/db/vpn/test_vpn_db.py:FLAVOR_PLUGIN_KLASS = "neutron.services.flavors.flavors_plugin.FlavorsPlugin"
unit/db/vpn/test_vpn_db.py:    def setUp(self, core_plugin=None, vpnaas_plugin=DB_VPN_PLUGIN_KLASS,
unit/db/vpn/test_vpn_db.py:            'flavors_plugin': FLAVOR_PLUGIN_KLASS}
```

asr1k_neutron_l3/tests/common/fixtures.py
sebageek: Yes, with this the vpn driver is always present. As we're going to have it enabled in all regions, I didn't bother writing an on/off switch for it, though we could think about doing that, just to be on the safe side. Might include a bit more effort though, not sure if it is worth it. Probably needs a discussion.

asr1k_neutron_l3/tests/common/flavor.py
sebageek: This is code I moved when refactoring the tests a bit, so I didn't touch that. The reason though is probably because the first block of ifs handles traits (required and optional), the second "block" is for quota handling, which was added separately.

asr1k_neutron_l3/tests/common/vpnaas.py
sebageek: The original file is [test_vpn_db.py](https://github.com/sapcc/neutron-vpnaas/blob/stable/2024.1-m3/neutron_vpnaas/tests/unit/db/vpn/test_vpn_db.py) from neutron-vpnaas. I have taken all create/delete methods from there and slightly modified them. Originally I hoped that I could inherit this class, but I would need to modify many places, as they use `self.api_ext` and I need to use `self.api` for most things to properly work. I also added extra arguments to certain create calls. 80% or the code is from that file though, so I think it is right to keep the copyright from them and then add our own to the list.

asr1k_neutron_l3/tests/common/vpnaas.py
sebageek: In these cases I just kept the original code from upstream and didn't review it. I planned on just taking it "as is". When rewriting it I'd probably do a couple of things different. What do you think, should I do these cleanup changes or do we keep it as is? Same for the following comments.

asr1k_neutron_l3/tests/common/vpnaas.py
sebageek: Same as above + I like to generally avoid overshadowing internal methods via function args if possible

asr1k_neutron_l3/extensions/vpn_asr1k_int_tunnel_ips.py
sebageek: There is a lot of validation that we do afterwards, so I don't know if a comment that says "this field needs further validation" is needed. In my mind it's "just" the API def with some preliminary validation and rest is done elsewhere.

I've written a test and some validation code that I'll push up with the first force push.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+        self.validate_internal_tunnel_ips(context, vpnservice_id)
+
+    def _autoallocate_int_tun_ips(self, context, tun_ips, ipsec_sitecon_id, vpnservice_id):
+        all_vpn_tun_ips = None
+        for ip_ver, def_subnet, def_prefixlen in ((4, self.INT_IPV4_NET, 30),  (6, self.INT_IPV6_NET, 126)):
+            net_key = f"local_cidr_v{ip_ver}"
+            ip_key = f"peer_address_v{ip_ver}"
+
+            # no auto alloc neede if both are specified
```
sven-rosenzweig: typo `neede`

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+
+        vpnservice_id = ipsec_sitecon_new['vpnservice_id']
+        tun_ips = self._autoallocate_int_tun_ips(context, tun_ips, ipsec_sitecon_id, vpnservice_id)
+        self.l3db.create_or_update_vpn_internal_tunnel_ips(context, ipsec_sitecon_id,
+                                                           tun_ips['local_cidr_v4'], tun_ips['peer_address_v4'],
+                                                           tun_ips['local_cidr_v6'], tun_ips['peer_address_v6'])
+
+        self.validate_internal_tunnel_ips(context, vpnservice_id)
+
+    def _autoallocate_int_tun_ips(self, context, tun_ips, ipsec_sitecon_id, vpnservice_id):
```
sven-rosenzweig: Do we need `ipsec_sitecon_id` as a parameter here? Do not see any usage.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+        tun_ips = self.l3db.get_vpn_internal_tunnel_ips(context, vpnservice_id=vpnservice_id)
+
+        for tun_ip in tun_ips:
+            tun_ip['local_cidr_v4'] = ipaddress.ip_interface(tun_ip['local_cidr_v4'])
+            tun_ip['local_cidr_v6'] = ipaddress.ip_interface(tun_ip['local_cidr_v6'])
+            tun_ip['peer_address_v4'] = ipaddress.ip_address(tun_ip['peer_address_v4'])
+            tun_ip['peer_address_v6'] = ipaddress.ip_address(tun_ip['peer_address_v6'])
+
+            for ip_ver, max_prefix_len_for_net_addr in ((4, 31), (6, 127)):
+                local_net = tun_ip[f'local_cidr_v{ip_ver}'] = ipaddress.ip_interface(tun_ip[f'local_cidr_v{ip_ver}'])
```
sven-rosenzweig: Could we do: 
`local_net = tun_ip[f'local_cidr_v{ip_ver}'] = ipaddress.ip_interface(tun_ip[f'local_cidr_v{ip_ver}'])`
--> 
`local_net = tun_ip[f'local_cidr_v{ip_ver}']`

as in line 239 and 240, we already store IPv4Interface/IPv46Interface datatypes in our tun_ip dict?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
```diff
@@ -0,0 +1,429 @@
...
+            if prefixes := peer_cidrs & other_peer_cidrs:
+                raise DuplicateEndpointsBetweenIPSecSiteConnections(
+                        ipsec_site_connection_id=sitecon['id'],
+                        other_ipsec_site_connection_id=other_sitecon['id'],
+                        prefixes=", ".join(prefixes))
+
+
+class ASR1KIPSecVPNaaSNotifier:
+    def vpnservice_updated(self, context, router_id, **kwargs):
+        # handled via hooks
```
sven-rosenzweig: Reading this the question popped into my mind why we implement this via hooks. Maybe we could add a short reasoning here?

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: It looks like both of these checks are refactoring leftovers, I'll remove them.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: I think it's not important why we update the router. In most cases it will be create or delete, as most of the update operations don't properly work with neutron-vpnaas.

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: I'll remove the create as discussed, not really needed

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: Don't need it there, I'll remove it

asr1k_neutron_l3/neutron/services/service_drivers/asr1k/vpnaas_driver.py
sebageek: We need it to handle an ipsec siteconn delete differently, I can add a comment regarding it.

asr1k_neutron_l3/models/neutron/l3/router.py
sebageek: We've tried a bit to come up with a better solution and... didn't come up with something that we liked. I think we're gonna leave this for another time.

asr1k_neutron_l3/plugins/l3/agents/device_cleaner.py
sebageek: I don't think they're ever triggered, as nearly nobody knows about the API and none of our tools use it. The driver in normal operation doesn't use it as well.

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -186,9 +186,24 @@ def get_network_port_count_per_agent(self, context: n_context.Context, network_i
...
+            LOG.debug("Setting status to %s for router %s", status, router_id)
             router = {'router': {'status': status}}
             ctx = n_context.get_admin_context()
             self.update_router(ctx, router_id, router)
+
+            if utils.is_vpnaas_enabled():
+                vpn_plugin = directory.get_plugin("VPN")
+                sitecon_ids = self.get_ipsec_site_connection_ids(ctx, router_id=router_id)
+                if sitecon_ids:
+                    first_sitecon = vpn_plugin.get_ipsec_site_connection(ctx, sitecon_ids[0])
```
sven-rosenzweig: [nit]: We need to fetch the first sitecon in order to get the vpnservice_id, right? Stumbling over this line I was asking myself, we wee need the first. 

Could we do:
``vpnservice_id = vpn_plugin.get_ipsec_site_connection(ctx, sitecon_ids[0]).get('vpnservice_id')``

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -186,9 +186,24 @@ def get_network_port_count_per_agent(self, context: n_context.Context, network_i
...
+                    for sitecon_id in sitecon_ids:
+                        LOG.debug("Setting status to %s for router %s ipsec site connection %s",
+                                  status, router_id, sitecon_id)
+                        vpn_plugin.update_ipsec_site_conn_status(ctx, sitecon_id, status)
+                    LOG.debug("Setting status to %s for router %s vpnservice %s",
+                              status, router_id, first_sitecon['vpnservice_id'])
+                    vpn_plugin.set_vpnservice_status(ctx, first_sitecon['vpnservice_id'],
+                                                     status, updated_pending_status=True)
         except l3_exc.RouterNotFound:
```
sven-rosenzweig, line 193: Wondering if we need to catch vpnservice related exceptions here? Looking at ``vpn_plugin.get_ipsec_site_connection(ctx, sitecon_ids[0])`` a ``vpn_exception.IPsecSiteConnectionNotFound``  migt be raised deeper down the code.

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -703,6 +718,108 @@ def get_router_count_by_flavor_name(self, context, flavor_name, project_id=None)
...
+                'local_cidr_v4': entry.local_cidr_v4,
+                'peer_address_v4': entry.peer_address_v4,
+                'local_cidr_v6': entry.local_cidr_v6,
+                'peer_address_v6': entry.peer_address_v6,
+            }
+            for entry in query
+        ]
+
+    @db_api.CONTEXT_WRITER
+    def create_or_update_vpn_internal_tunnel_ips(self, context, ipsec_sitecon_id,
```
sven-rosenzweig, line 744: Is there a way to determine if a ``create``  or ``update`` is needed without asking the database? I am wondering if we cloud safe one call to the DB in case of a create action. 

Although I have to admit it is very convenient to handle create and update actions in the same call. Makes the code readable.

asr1k_neutron_l3/plugins/db/asr1k_db.py
```diff
@@ -703,6 +718,108 @@ def get_router_count_by_flavor_name(self, context, flavor_name, project_id=None)
...
+                                                 local_cidr_v4, peer_address_v4, local_cidr_v6, peer_address_v6):
+        query = context.session.query(asr1k_models.ASR1KInternalTunnelIp)
+        query = query.filter(asr1k_models.ASR1KInternalTunnelIp.ipsec_site_connection_id == ipsec_sitecon_id)
+        if query.count() > 0:
```
sven-rosenzweig, line 748: [nit]: As discussed in Slack: ``ASR1KInternalTunnelIp`` to ``IPSecSeitConnections`` is a one to one relationship. Not knowing this for sure, I was a little bit confused by the if/else statement. 
Maybe a comment could clarify this.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
```diff
@@ -353,8 +368,45 @@ def get_sync_data(self, context, router_ids=None, active=None, host=None):
                 all_port_ids = [port["id"] for port in all_ports]
                 router["fwaas_policies"] = self.get_fwaas_policies(context, all_port_ids)
 
+            router["vpn"] = None
+            if utils.is_vpnaas_enabled() and router["flavor_id"]:
```
sven-rosenzweig, line 372: [nit] We could store this in a variable ``vpnaas_enabled = utils.is_vpnaas_enabled()`` in the constructor. In another method of this class the same function was called.

sebageek: We could also let utils.is_vpnaas_enabled() method cache that, what do you think?

asr1k_neutron_l3/tests/common/flavor.py
```diff
@@ -0,0 +1,81 @@
...
+    def _make_service_profile(self, ctx, metainfo, driver=None, enabled=True, description=''):
+        sp_def = {'service_profile': {
+            'driver': driver or self.node_driver,
+            'enabled': enabled,
+            'metainfo': metainfo,
+            'description': description,
+        }}
+        return self.fp.create_service_profile(ctx, sp_def)
+
+    def _make_agent(self, ctx, host, configuration=None, az='nova'):
```
sven-rosenzweig, line 47: ``ctx`` is not used.

asr1k_neutron_l3/tests/unit/services/service_drivers/asr1k/test_vpnaas_driver.py
```diff
@@ -0,0 +1,338 @@
...
+        vpn2 = self._create_vpnservice("json", "vpn1", True, router['router']['id'], None, as_admin=True)
+        self.assertEqual("MultipleVPNServicesOnRouterDisallowed", vpn2['NeutronError']['type'])
+        self.assertIn(vpn1['vpnservice']['id'], vpn2['NeutronError']['message'])
+
+    def test_router_with_cipher_checks(self):
+        router = self._make_vpn_ready_router()
+        vpn = self._create_vpnservice("json", "vpn1", True, router['router']['id'], None, as_admin=True)
+
+        sc_objs = self._make_sitecon_related_objs(ike_args={"ike_version": "v1"})
+        sitecon = self._create_ipsec_site_connection("json",
```
sven-rosenzweig: ``sitecon`` variable is not used.

asr1k_neutron_l3/plugins/l3/service_plugins/l3_extension_adapter.py
sven-rosenzweig: I like this method ``utils.is_vpnaas_enabled()``. Closing conversation.

asr1k_neutron_l3/tests/common/flavor.py
mutax: where? its used in the return statement?

asr1k_neutron_l3/tests/common/vpnaas.py
mutax: hm yea, would need a different name...  
general discussion above at the first mention of that.

asr1k_neutron_l3/tests/common/vpnaas.py
mutax: I am fine with both. When writing new code I am in favor of being able to see directly which parameters are valid for a method, where possible. In general I am not a huge fan of (over) using the kwargs/args pattern. But I don't feel we need a discussion at this point here for that code. Fine with staying with upstream

asr1k_neutron_l3/tests/common/flavor.py
sebageek: I fixed that in [this force push](https://github.com/sapcc/asr1k-neutron-l3/compare/4c47a6a61a6a6d9e6dcc1df76333f10ff5285d45..730847d103b2b94a8db71f4a34b69773396eaa3e#diff-efae66eceafab1bb28d1c019625f78d70a60d66b05f83e8027ebe02f3cdfcaf2L58-L59)

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: As discussed in call, moved get_ipsec_site_connection() to the block where it makes a bit more sense.

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: Moving the whole vpn setting stuff outside of the try/except, as it is unrelated to it. As we're fetching everything from the DB I'd expect us not to get exceptions here regularly. If we do, the exception should fly unhandled.

asr1k_neutron_l3/plugins/db/asr1k_db.py
sebageek: I went for this due to convenience. Normally this is done during creation, but if anything goes wrong we'll get the row also in an update case. "Just make sure it's there".

sven-rosenzweig (approved):

mutax (approved):

PR #144, opened 2025-10-31
Deduplicate ACEs before creating rest definition

swagner-de:
Openstack happily accepts duplicate entries, the device however doesn't.
We will continue to allow users to create duplicate entries, yet chose
to dedeuplicate them before pushing to the device.

### DIFF
--- a/asr1k_neutron_l3/models/neutron/l3/access_list.py
+++ b/asr1k_neutron_l3/models/neutron/l3/access_list.py
@@ -27,7 +27,14 @@ class AccessList(base.Base):
     @property
     def _rest_definition(self):
         acl = access_list.AccessList(name=self.id)
-        for i, rule in enumerate(self.rules):
+        rule_set = set()
+        dedup_rules = []
+        for rule in self.rules:
+            if rule in rule_set:
+                continue
+            dedup_rules.append(rule)
+            rule_set.add(rule)
+        for i, rule in enumerate(dedup_rules):
             sequence = (i + 1) * 10
             ace_rule = access_list.ACERule(
                 access_list=self.id,
@@ -62,12 +69,30 @@ class Rule():
         self.protocol = protocol
         self.source = source
         self.source_mask = source_mask
-        self.source_port_range = source_port_range
+        # Make port range a tuple for easier hashing and equality checks
+        self.source_port_range = tuple(source_port_range) if source_port_range else None
         self.destination = destination
         self.destination_mask = destination_mask
-        self.destination_port_range = destination_port_range
+        # Make port range a tuple for easier hashing and equality checks
+        self.destination_port_range = tuple(destination_port_range) if destination_port_range else None
         self.named_message_type = named_message_type
         self.established = established
+
+    def _get_tuple(self):
+        """Return tuple of all fields used for hashing and equality."""
+        return (self.action, self.protocol,
+                self.source, self.source_mask, self.source_port_range,
+                self.destination, self.destination_mask,
+                self.destination_port_range,
+                self.named_message_type, self.established)
+
+    def __hash__(self):
+        return hash(self._get_tuple())
+
+    def __eq__(self, value):
+        if not isinstance(value, Rule):
+            return False
+        return self._get_tuple() == value._get_tuple()
 
     @property
     def ip_args(self):
--- a/asr1k_neutron_l3/tests/unit/models/neutron/l3/test_access_list.py
+++ b/asr1k_neutron_l3/tests/unit/models/neutron/l3/test_access_list.py
@@ -1,0 +1,53 @@
+# Copyright 2026 SAP SE
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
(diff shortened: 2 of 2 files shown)


felix-kaestner (dismissed):

asr1k_neutron_l3/models/neutron/l3/access_list.py
```diff
@@ -69,6 +76,12 @@ def __init__(self, action='permit', protocol='ip',
         self.named_message_type = named_message_type
         self.established = established
 
+    def __hash__(self):
```
sebageek, line 89: If we implement `__hash__()` we also need to implement `__eq__()`, see [0]. If we only implement `__hash__()`, python finds the right hash bucket but then in the bucket cannot properly test for equality.

[0] https://docs.python.org/3/reference/datamodel.html#object.__hash__

sebageek (changes requested):
A test would be nice, but I'd rather have this feature without a test than not have it. If you need/want support in writing a short test for the behavior I'd also be open to support.

asr1k_neutron_l3/models/neutron/l3/access_list.py
```diff
@@ -69,6 +76,12 @@ def __init__(self, action='permit', protocol='ip',
...
+        return hash((self.action, self.protocol,
+                     self.source, self.source_mask, self.source_port_range,
+                     self.destination, self.destination_mask, self.destination_port_range,
```
sebageek: Are we sure `source_port_range` and `destination_port_range` are never a list? We could just call `tupe()` on it here to make sure it's always hashable (which `list()` is not). Just to be future proof™.

sebageek:
Here's a suggestion for a test in `asr1k_neutron_l3/tests/unit/models/neutron/l3/test_acess_list.py`:

```python
# [... license header ...]
from asr1k_neutron_l3.models.neutron.l3.access_list import AccessList, Rule
from asr1k_neutron_l3.tests.common.fixtures import RouterWithSyncDataTestCase


class TestAccessList(RouterWithSyncDataTestCase):
    def test_access_list_rule_deduplication(self):
        acl = AccessList(id='test-acl')

        # create two identical rules (must be two rules, so they are different instances, for eq check)
        # TODO: change destination_port_range to list if we do the tuple conversion in `__hash__()`
        rule1 = Rule(action='permit', protocol='tcp', source='10.0.0.1', destination='10.0.0.2', destination_port_range=("443",))
        rule2 = Rule(action='permit', protocol='tcp', source='10.0.0.1', destination='10.0.0.2', destination_port_range=("443",))

        acl.append_rule(rule1)
        acl.append_rule(rule2)

        self.assertEqual(1, len(acl._rest_definition.rules))

```

asr1k_neutron_l3/models/neutron/l3/access_list.py
swagner-de: Makes sense

asr1k_neutron_l3/models/neutron/l3/access_list.py
swagner-de: So far the only code part I could find where we set `Rule.(source|destination)_port_range` is [here](https://github.com/sapcc/asr1k-neutron-l3/blob/dedup-acl/asr1k_neutron_l3/models/neutron/l3/firewall.py#L100-L102) and that one uses a `tuple()`.
How about I add a type hint?

asr1k_neutron_l3/tests/unit/models/neutron/l3/test_acess_list.py
```diff
@@ -0,0 +1,54 @@
...
+from asr1k_neutron_l3.tests.common.fixtures import RouterWithSyncDataTestCase
+
+
+class TestAccessList(RouterWithSyncDataTestCase):
+
+    def test_access_list_rule_deduplication(self):
+        acl = AccessList(id='test-acl')
+
+        # create two identical rules (must be two rules, so they are different instances, for eq check)
+        # TODO: change destination_port_range to list if we do the tuple conversion in `__hash__()`
```
sebageek: That todo is done now :)

asr1k_neutron_l3/models/neutron/l3/access_list.py
```diff
@@ -62,13 +69,33 @@ def __init__(self, action='permit', protocol='ip',
...
+                     self.named_message_type, self.established))
+
+    def __eq__(self, value):
+        if not isinstance(value, Rule):
+            return False
+        return (self.action, self.protocol,
+                self.source, self.source_mask, self.source_port_range,
+                self.destination, self.destination_mask, self.destination_port_range,
+                self.named_message_type, self.established) == \
```
sebageek: [optional] As we have this tuple now used three times (once in `__hash__()` and two times in `__eq__()`) we could put this into a `def _get_tuple(self) -> tuple:` method or something so we have all the values in one place.

sebageek (dismissed):
Typo in `test_acess_list.py`, should be access_list, otherwise awesome :)

sebageek (approved):

felix-kaestner (approved):

sven-rosenzweig (approved):

PR #146, opened 2025-12-01
Sort routes by prefix + netmask for better diff

sebageek:
Whenever we update something on the device, we check if the config is already present by doing a diff (an operation also used by admins to check if the config has been synced properly to the hardware router). In some cases we now have the same config on the device and in OpenStack, but the ordering is wrong (as Cisco and Neutron have their separate ordering). In this case we have this problem for static routes, where routes with different netmasks are sorted differently.

### DIFF
--- a/asr1k_neutron_l3/models/netconf_yang/route.py
+++ b/asr1k_neutron_l3/models/netconf_yang/route.py
@@ -104,7 +104,7 @@ class VrfRouteBase(NyBase):
             return {}
 
         vrf_routes = []
-        for route in sorted(self.routes, key=lambda route: route.prefix):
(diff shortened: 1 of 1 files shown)


toanju (approved):
lgtm, should there be tests for this?

sebageek:
There is two things that we could test here:
 * that it still works (i.e. no exception on calling the `to_dict()` method
 * that there is a specific order
For the first thing I'd like #143 to be merged, as this reworks some parts of the testing framework. For the second test: I don't know if it's necessary, went for no, but would be open for also writing them.

sven-rosenzweig (approved):

PR #161, opened 2026-05-05
Fix alembic migrations for initial deployment

sebageek:
With the current alembic migrations that we have, initial deployment does not work. This fixes this and cleans up the expand/contract migrations. This will break backwards compatibility with regions that are not yet fully migrated to a expand db version before  `fe1f53f6a61d`. Please see the individual commits for the full details.


occamshatchet (approved):
lgtm

swagner-de (approved):

